PR #27, opened 2018-02-20
Fixed problem starting tsserver from local node_modules on windows (see #2)

svenefftinge:


src/utils.ts
```diff
@@ -1,3 +1,5 @@
+import { clearTimeout } from "timers";
```
akosyakov: imports in the bogus place

src/non-file-uri.spec.ts
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (C) 2017 TypeFox and others.
```
akosyakov: 2018

PR #36, opened 2018-04-18
Clear diagnostics when closing file

simark:
When the user closes a file which has some errors/warnings, we currently
leave those diagnostics in place.  However, this language server doesn't
provide diagnostic for closed files.  This means that the diagnostics
won't get updated until the user re-opens that file.  They can therefore
get stale, if the error or warning condition disappears while the file
is closed.

This patch makes it so that when a file is closed, we clear the
diagnostics associated to that file.

Signed-off-by: Simon Marchi <simon.marchi@ericsson.com>


simark:
Thanks, you'll have to merge it yourself though, I don't have write access.  I also suppose we'll need a release in order to be able to use it in Theia.

PR #44, opened 2018-05-22
Fix dependency vulnerabilities

zewa666:
Your project is configured to run via yarn, which sadly does not implement the audit feature like npm yet. While trying to get my hands dirty with the repo I figured that there was a critical dependency issue for `command-exists` which is a base dependency, thus all consumers would inherit this issue as well. On that take I've also updated `mocha` to the most recent version to get rid of all vulnerability warnings.

audit-results

Since mocha >= v4 has a different behavior of force killing the process, the last commit adds the `--exit` behavior to act in pre v4 mode. This typically indicates that there are unresolved tests, which run forever and do not properly quit.

Last but not least, my VSCode setup went crazy without a proper editorconfig, so I've created one to my best second guess efforts. Most of the other Theia projects follow a different pattern (2 spaces, forced newline at the end of files) but files in here seem to be using varying settings. Might make sense to run a chore PR and get all straight to the same base settings.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = false
+insert_final_newline = false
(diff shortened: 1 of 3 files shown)


zewa666:
Alright it seems that the `--exit` flag was sufficient to please Travis

svenefftinge:
thanks

PR #56, opened 2018-08-21
apply rename file

akosyakov:
This PR adds a command to apply edits on a file rename. I'm going to test it with Theia.


akosyakov:
Verified that it works with ts 3.0 and higher: https://github.com/theia-ide/theia/pull/2629#issue-209769782

PR #59, opened 2018-08-31
allow a client to enable tsserver logging

akosyakov:
It is needed to allow clients to track down misbehaving tsservers.

I will open a PR with Theia to try it out.


akosyakov:
A PR to try out with Theia is here: https://github.com/theia-ide/theia/pull/2742

Issue #70, opened 2018-09-12
When multiple files are open, only diagnostics from the last accessed file are updated


keyboardDrummer:
The cause of this bug is that [this call to debounce](https://github.com/theia-ide/typescript-language-server/blob/master/server/src/diagnostic-queue.ts#L36) let's diagnostics for one file debounce diagnostics for another file, while the purpose of using debounce here is only to aggregate diagnostics of different kinds within a single file, into a single notification.

keyboardDrummer:
PR [here](https://github.com/theia-ide/typescript-language-server/pull/71)

PR #71, opened 2018-09-12
Publish diagnostics for all open files when multiple files are open

keyboardDrummer:
### Notes
- We now create a `debounce` function per open file, so two different files will not 'bounce' the diagnostics for one another.
- I had to use test specific uri's in the tests to prevent the tests from interacting. Even though the tests run in series, they re-use some state (the LSP server) which will causes the tests to interact. Not re-using the state is, I think, not an option because it will slow down the tests.

### Testing
- Ran the tests
- Tested against an IDE. This CR resolves [this issue](https://github.com/theia-ide/typescript-language-server/issues/70)

### DIFF
--- a/server/src/diagnostic-queue.ts
+++ b/server/src/diagnostic-queue.ts
@@ -12,12 +12,36 @@ import { pathToUri, toDiagnostic } from './protocol-translation';
 import { EventTypes } from './tsp-command-types';
 import debounce = require('p-debounce');
 
+class FileDiagnostics {
+    private readonly diagnosticsPerKind = new Map<EventTypes, lsp.Diagnostic[]>();
+    
+    constructor(readonly uri: string) { }
+    
+    private readonly returnDiagnostics = debounce((resolve: (params: lsp.PublishDiagnosticsParams) => void) => {
+        const diagnostics = this.getDiagnostics();
+        resolve({ uri: this.uri, diagnostics });
+    }, 50);
+    
+    public updateDiagnostics(kind: EventTypes, diagnostics: tsp.Diagnostic[]): Promise<lsp.PublishDiagnosticsParams> {
+        this.diagnosticsPerKind.set(kind, diagnostics.map(toDiagnostic));
+        return new Promise((resolve) => this.returnDiagnostics(resolve));
+    }
+    
+    protected getDiagnostics(): lsp.Diagnostic[] {
+        const result: lsp.Diagnostic[] = [];
+        for (const value of this.diagnosticsPerKind.values()) {
+            result.push(...value);
+        }
+        return result;
+    }
+}
+
 export class DiagnosticEventQueue {
 
-    protected readonly diagnostics = new Map<string, Map<EventTypes, lsp.Diagnostic[]>>();
+    protected readonly diagnostics = new Map<string, FileDiagnostics>();
 
     constructor(
-        protected publicDiagnostics: (params: lsp.PublishDiagnosticsParams) => void,
+        protected publishDiagnostics: (params: lsp.PublishDiagnosticsParams) => void,
         protected logger: Logger
     ) { }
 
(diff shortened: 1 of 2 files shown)


server/src/diagnostic-queue.ts
```diff
@@ -27,27 +51,11 @@ export class DiagnosticEventQueue {
...
-        const result: lsp.Diagnostic[] = [];
-        for (const value of diagnostics.values()) {
-            result.push(...value);
+        let fileDiagnostics = this.diagnostics.get(file);
+        if (!fileDiagnostics) {
+            fileDiagnostics = new FileDiagnostics(pathToUri(file));
+            this.diagnostics.set(file, fileDiagnostics);
         }
-        return result;
+        fileDiagnostics.updateDiagnostics(kind, event.body.diagnostics).then(this.publishDiagnostics);
```
akosyakov: Won't it execute `publishDiagnostics` so many times as `updateDiagnostics` is called instead of once? How about passing `publishDiagnostics` to `FileDiagnostics` and using `firePublishDiagnostics` as it was before, i.e. without any async logic.

akosyakov:
@keyboardDrummer Goot catch!

Before it will postpone publishing diagnostics for all files while tsserver is spamming with diagnostic events, e.g. during the full build on a monorepo. With this change, diagnostics debounced only for a single file leading to republishing a lot of unnecessary diagnostics in some cases. I would prefer that we debounce for all files as before, but fix an issue with publishing only for last.

server/src/diagnostic-queue.ts
keyboardDrummer: > Won't it execute `publishDiagnostics` as many times as `updateDiagnostics` is called instead of once?

It won't. If you look at `updateDiagnostics`, you can see it only resolves the promise once if multiple calls happen within the debounce window, since the call to `resolve` is within the function passed to `debounce`.

keyboardDrummer:
> Before it will postpone publishing diagnostics for all files while tsserver is spamming with diagnostic events, e.g. during the full build on a monorepo. With this change, diagnostics debounced only for a single file leading to republishing a lot of unnecessary diagnostics in some cases. I would prefer that we debounce for all files as before, but fix an issue with publishing only for last.

The current implementation will publish diagnostics for all open files whenever diagnostics change, this allows IDE's to implement something like a problems view, where diagnostics are shown not just for the last edited file but for multiple files.

If you look at the problems in VS Code you will see that it shows errors for all open files (for TypeScript).

akosyakov:
> The current implementation will publish diagnostics for all open files whenever diagnostics change, this allows IDE's to implement something like a problems view, where diagnostics are shown not just for the last edited file but for multiple files.

I did not argue with that. Publishing diagnostics only for the last file is bogus. I've glad that you found and fixed it.

I said that I think that we should debounce publishing diagnostics for all files even by a single file to avoid posting unnecessary intermediate diagnostics during the full builds. Imagine you build several packages with complex interconnections, you will get all kind of intermediate diagnostics during the guild. Before this change only final diagnostics were supposed to be published, now there is a higher chance to get an intermediate diagnostics.

server/src/diagnostic-queue.ts
akosyakov: ok, although the code will be less tricky (without any async handling) on impl and client sides:
```ts
    constructor(
        readonly uri: string,
        protected publishDiagnostics: (params: lsp.PublishDiagnosticsParams) => void
    ) { }

    updateDiagnostics(kind: EventTypes, diagnostics: tsp.Diagnostic[]): void {
        this.diagnosticsPerKind.set(kind, diagnostics.map(toDiagnostic));
        this.firePublishDiagnostics();
    }
    protected firePublishDiagnostics = debounce((): void => {
        const { uri } = this;
        const diagnostics = this.getDiagnostics();
        this.publishDiagnostics({ uri, diagnostics });
    }, 50);
```

akosyakov:
@keyboardDrummer please take care of build failures and let's merge it

PR #77, opened 2018-09-20
Let documentSymbol return the correct results when mergeable elements are used

keyboardDrummer:
Previously, the documentSymbol results would grow exponentially in size when using multiple elements with the same name such as in some of the examples [for declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).

Note that this implementation, although in line with the existing implementation, is different from what VSCode does. An example, given the code:

```
interface Box {
    height: number;
    width: number;
}

interface Box {
    scale: number;
}
```

VSCode will show:
```
Box
  height
  width
  scale
```

While this implementation shows:
```
Box
  height
  width
Box
  scale
```

### DIFF
--- a/server/src/document-symbol.ts
+++ b/server/src/document-symbol.ts
@@ -7,19 +7,27 @@
 
 import * as lsp from 'vscode-languageserver';
 import * as tsp from 'typescript/lib/protocol';
-import { asRange, toSymbolKind, Range, toPosition } from "./protocol-translation";
+import { asRange, toSymbolKind, Range } from "./protocol-translation";
 import { ScriptElementKind } from './tsp-command-types';
 
 export function collectDocumentSymbols(parent: tsp.NavigationTree, symbols: lsp.DocumentSymbol[]): boolean {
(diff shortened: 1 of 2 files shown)


akosyakov:
Can you elaborate what was the issue? I cannot reproduce it.

I've tried the master state and latest VS Code and they both show:
```
Box
  height
  width
Box
  scale
```
screen shot 2018-09-20 at 17 59 23

master_document_symbols

server/src/document-symbol.ts
```diff
@@ -40,7 +48,7 @@ export function collectDocumentSymbols(parent: tsp.NavigationTree, symbols: lsp.
 }
 
 export function collectSymbolInformations(uri: string, current: tsp.NavigationTree, symbols: lsp.SymbolInformation[], containerName?: string): boolean {
```
akosyakov: We should probably do the same here for clients who cannot handle hierarchical symbols, can be a separate PR.

PR #78, opened 2018-09-21
Spawn TSServer from same Node.js executable as wrapper

mattlyons0:
Currently `spawn` is being called on the `.bin/tsserver` file: https://github.com/Microsoft/TypeScript/blob/master/bin/tsserver

This means the hashbang at the top is being followed to launch TSServer with `/usr/bin/env node`. This executable may be different than the executable used to launch the Theia wrapper. `process.execPath` will get the path of the executable running the wrapper and execute TSServer with that.

My use case for this is running Theia as an Atom plugin and using electron's built in node executable instead of the globally installed one. This solves the issue recently discussed in https://github.com/atom/ide-typescript/issues/127 (and probably alleviates potential future headaches if someone has multiple node versions installed)

### DIFF
--- a/server/src/lsp-server.ts
+++ b/server/src/lsp-server.ts
@@ -71,18 +71,33 @@ export class LspServer {
     }
 
     protected findTsserverPath(): string {
-        if (this.options.tsserverPath) {
+        if (this.options.tsserverPath) { // Attempt to maintain backwards compatability
+          if (fs.existsSync(this.options.tsserverPath)) {
             return this.options.tsserverPath;
+          } else {
+            try {
+              return require.resolve(this.options.tsserverPath)
+            } catch (e) {
+              try {
+                return require.resolve('typescript/bin/' + this.options.tsserverPath)
+              } catch (e) {
+                throw Error(`Could not find passed TSServer Path: '${this.options.tsserverPath}'. Path does not exist nor does resolve to a module.`)
+              }
+            }
+          }
         }
+
         // 1) look into node_modules of workspace root
         let executable = findPathToModule(this.rootPath(), `.bin/${getTsserverExecutable()}`)
         if (executable) {
             return executable;
         }
         // 2) use globally installed tsserver
-        if (commandExists.sync(getTsserverExecutable())) {
-            return getTsserverExecutable();
-        }
+        try {
+          let path = require.resolve(getTsserverExecutable())
+          return path
+        } catch (e) {}
+
         // 3) look into node_modules of typescript-language-server
         const bundled = findPathToModule(__dirname, `.bin/${getTsserverExecutable()}`);
         if (!bundled) {
--- a/server/src/tsp-client.ts
+++ b/server/src/tsp-client.ts
@@ -86,7 +86,7 @@ export class TspClient {
             return;
         }
         const { tsserverPath, logFile, logVerbosity, globalPlugins, pluginProbeLocations } = this.options;
-        const args: string[] = []
+        const args: string[] = [ tsserverPath ]
         if (logFile) {
             args.push('--logFile', logFile);
         }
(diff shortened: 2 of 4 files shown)


mattlyons0:
It seems a side effect of this PR is that the way the file was being resolved in your tests no longer works. 

I don't actually understand why that even worked before, is it something special to do with the way the tests are written/configured or is it something users could do as well. I'm wondering if this PR is a breaking change or not (since you allow users to specify the tsserver and it seems the accepted path types has at least changed a little bit).

akosyakov:
It seems to be reasonable. But the issue is that before `tsserverPath` actually was pointing to an executable, now it should be an absolute path to typescript location and it breaks here: https://github.com/theia-ide/typescript-language-server/blob/1d70a4271f8e5ef688f744630da9735835de1564/server/src/lsp-server.ts#L73-L92

We will need to rewrite this method that it returns a proper absolute path for all cases. Also instead of spawning binaries, we can directly spawn `typescript/lib/tsserver.js`.

And to spawn a similar Node.js one should just use `cp.fork`.

mattlyons0:
Sorry about dropping the ball on this PR, I have some fixes coming soon

server/src/lsp-server.ts
```diff
@@ -80,9 +80,11 @@ export class LspServer {
             return executable;
         }
         // 2) use globally installed tsserver
-        if (commandExists.sync(getTsserverExecutable())) {
-            return getTsserverExecutable();
-        }
+        try {
+          let path = require.resolve(getTsserverExecutable())
+          return path
+        } catch (e) {}
```
mattlyons0: Before just `tsserver` was being executed since we had the ability to execute files (which would start with an arbitrary node server, whatever the hashbang resolved to). Now we resolve the package `typescript/bin/tsserver` to get the path on disk.

server/src/tsp-client.ts
```diff
@@ -102,7 +102,7 @@ export class TspClient {
         this.cancellationPipeName = tempy.file({ name: 'tscancellation' } as any);
         args.push('--cancellationPipeName', this.cancellationPipeName + '*');
         this.logger.info(`Starting tsserver : '${tsserverPath} ${args.join(' ')}'`);
-        this.tsserverProc = cp.spawn(tsserverPath, args);
+        this.tsserverProc = cp.spawn(process.execPath, args);
```
mattlyons0: I'm not sure why but `cp.fork(tsserverPath, args)` (and changing args to not include the `tsserverPath` crashes on the next line (the readline interface creation). So I kept this as `cp.spawn`

server/src/utils.ts
```diff
@@ -33,7 +33,7 @@ export class Deferred<T> {
 }
 
 export function getTsserverExecutable(): string {
-    return isWindows() ? 'tsserver.cmd' : 'tsserver'
+    return isWindows() ? 'typescript/bin/tsserver.cmd' : 'typescript/bin/tsserver'
```
mattlyons0: I have only been testing this on Linux, since we are using `require.resolve` the resolver might take care of the windows edge case

mattlyons0:
Alright, so here is where we are. I can make this *mostly* backwards compatible but there is a case where it will not be. 

When you passed `--tsserver-path` before it would take anything executable. Now that behavior has changed. It will take anything executable by node (as long as the path can be resolved by node, this means relative and absolute paths will work but things resolved from your `$PATH` will not). Since things from `$PATH` are not resolved I am doing an additional check to see, 1. If I can `require()` the `tsserver-path` you specify OR 2. I can `require()` the `typescript/bin/tsserver-path`. So I'm thinking that the only case this breaks is if you were using `--tsserver-path` for something in your `$PATH` that wasn't from the `typescript` package. A way to fix that would be to call `which tsserver-path` and use that as the absolute path but that is getting rather hacky as it is system environment specific and such.

akosyakov:
@mattlyons0 changes in #88 seem to be enough for Atom, right? Could we close this PR?

mattlyons0:
Yes

PR #81, opened 2018-09-27
Return null when resolving completion request at an invalid location

keyboardDrummer:
Before this CR, asking for a completion request at an invalid location, such after the `b` in the code `function b` would cause the server to return an `ErrorCodes.InternalError`. This CR corrects that so the server just returns `null`, as is allowed by the LSP spec.

The reason for this is that tsserver, when you ask for completions at a nonsense location, returns a response with `success === false` ([ref](https://github.com/Microsoft/TypeScript/blob/0a97663843e11f745d9998d54db0970ee3bfb748/src/server/session.ts#L2364)). This LSP server will then reject the tsp request Promise ([ref](https://github.com/theia-ide/typescript-language-server/blob/master/server/src/tsp-client.ts#L197)), and this rejected Promise is never caught so it is later translated to an internal error.


server/src/lsp-server.ts
```diff
@@ -422,15 +422,24 @@ export class LspServer {
...
+                line: params.position.line + 1,
+                offset: params.position.character + 1,
+                includeExternalModuleExports: true,
+                includeInsertTextCompletions: true
+            }));
+            const body = result.body || [];
+            return body.map(entry => asCompletionItem(entry, file, params.position, document));
+        } catch (error) {
+            if (error.message === "No content available.") {
+                this.logger.warn('No content was available for completion request');
```
akosyakov: Do we need to log? Clients will see them in the output panel, they are more noise than helpful.

keyboardDrummer: Well, could be useful for debugging. Do you prefer we make it `this.logger.info` ?

PR #88, opened 2018-10-22
Allowing starting tsserver as a module using cp.fork

keyboardDrummer:
Previously, tsserver could only be started as an executable, which would often run the `.bin/tsserver` file that uses a Node shebang to start `tsserver.js` using the first Node on the `$PATH`.

With this PR, tsserver can also be started as a module, by setting `tsserverPath` so it points to `tsserver.js`. Also, when tsserver is bundled with the TS LSP server, it will be started using `cp.fork`, so it uses the same version of Node as the TS LSP server.

This PR is related to https://github.com/theia-ide/typescript-language-server/pull/78

### DIFF
--- a/server/src/tsp-client.ts
+++ b/server/src/tsp-client.ts
@@ -86,7 +86,7 @@ export class TspClient {
             return;
         }
         const { tsserverPath, logFile, logVerbosity, globalPlugins, pluginProbeLocations } = this.options;
-        const args: string[] = []
+        const args: string[] = [];
         if (logFile) {
             args.push('--logFile', logFile);
         }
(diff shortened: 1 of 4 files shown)


server/src/tsp-client.ts
```diff
@@ -102,7 +102,10 @@ export class TspClient {
         this.cancellationPipeName = tempy.file({ name: 'tscancellation' } as any);
         args.push('--cancellationPipeName', this.cancellationPipeName + '*');
         this.logger.info(`Starting tsserver : '${tsserverPath} ${args.join(' ')}'`);
-        this.tsserverProc = cp.spawn(tsserverPath, args);
+        const tsserverPathIsModule = path.extname(tsserverPath) === ".js";
```
mattlyons0: Maybe should support ".mjs" as well

Issue #53, opened 2018-08-14
Consider replacing timeout logic with appropriate request cancellation


akosyakov:
I'm getting following errors:
```
[Error - 00:15:15] Request textDocument/foldingRange failed.
Message: Request textDocument/foldingRange failed with message: getOutliningSpans timeout
Code: -32603
[Error - 00:15:16] Request textDocument/foldingRange failed.
Message: Request textDocument/foldingRange failed with message: getOutliningSpans timeout
Code: -32603
```

rchl:
The timeout was removed in b9cf57816ca07b9b3f3631130b7b5250b0db7ced

PR #92, opened 2018-11-15
Remove hardcoded request timeouts

### LINKED ISSUE #91: Make timeouts configurable / possible to disable
It seems like there is a hardcoded timeout for every request.
When running `yarn` while the server is running, it seems like tsserver gets overwhelmed with handling all the new files added and doesn't respond to requests immediatly.
I then get a lot of errors like:
```
Error: Request textDocument/definition failed with message: definition timeout
    at handleResponse (/Users/felix/src/github.com/sourcegraph/ext-lang-typescript/node_modules/vscode-jsonrpc/lib/main.js:436:48)
    at processMessageQueue (/Users/felix/src/github.com/sourcegraph/ext-lang-typescript/node_modules/vscode-jsonrpc/lib/main.js:263:17)
    at Immediate.setImmediate (/Users/felix/src/github.com/sourcegraph/ext-lang-typescript/node_modules/vscode-jsonrpc/lib/main.js:247:13)
```

felixfbecker:
Fixes #91 

I opted for removing the timeouts, because how much time a request should have to complete should really be determined by the client, and it should be possible to determine for each request or situation independently. And LSP already has a way to do this, which is `$/cancelRequest` - so when a request exceeded a timeout set by the client, the client can simply cancel the request.

I removed the functionality from `Deferred` since it is used no where else in the codebase but let me know if you would like me to keep it.

### DIFF
--- a/server/src/tsp-client.ts
+++ b/server/src/tsp-client.ts
@@ -133,7 +133,7 @@ export class TspClient {
     ): Promise<TypeScriptRequestTypes[K][1]> {
         this.sendMessage(command, false, args);
         const seq = this.seq;
-        const request = (this.deferreds[seq] = new Deferred<any>(command)).promise;
+        const request = (this.deferreds[seq] = new Deferred<any>()).promise;
         if (token) {
             const onCancelled = token.onCancellationRequested(() => {
                 onCancelled.dispose();
--- a/server/src/utils.ts
+++ b/server/src/utils.ts
@@ -9,26 +9,12 @@ import { clearTimeout } from "timers";
 
 export class Deferred<T> {
 
-    private timer: any
-
-    constructor(private operation: string, timeout?: number) {
-        this.timer = setTimeout(() => {
-            this.reject(new Error(this.operation + " timeout"));
-        }, timeout || 20000)
-    }
-
     resolve: (value?: T) => void;
     reject: (err?: any) => void;
 
     promise = new Promise<T>((resolve, reject) => {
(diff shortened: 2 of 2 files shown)


akosyakov:
I would like to get rid of timeouts as well: https://github.com/theia-ide/typescript-language-server/issues/53

But we should do it together with fixing handling of LSP cancellation tokens, right now they are completely ignored.

felixfbecker:
Oh, I wasn't aware of that. I would argue though that having no timeouts and no cancellation is still better than hardcoded timeouts.

felixfbecker:
@akosyakov are they really ignored? I see token handling here:

https://github.com/theia-ide/typescript-language-server/blob/a92027377b7ba8b1c9318baad98045e5128baa8e/server/src/tsp-client.ts#L140-L154

shackra:
Why was this abandoned?

harto:
This looks like a useful change to me. Does it need more work? I'm happy to help if I can.

harto:
Here's an updated commit with merge conflicts resolved: https://github.com/harto/typescript-language-server/commit/bbd64487c412e50bc8d76ed0c05b6e1a0ae55e28

DonnieWest:
@harto mind opening a PR with that? I'd love to give it a quick test and see if we can merge it

I need to look into our cancellation tokens too, make sure those are working as expected. Either way though, this looks mergeable

harto:
Sure: #189

Thanks!

PR #102, opened 2019-03-06
[type-hierarchy] Implement LSP proposal (wip)

AlexTugarev:
See: https://github.com/Microsoft/vscode-languageserver-node/pull/426

Signed-off-by: Alex Tugarev <alex.tugarev@typefox.io>

### DIFF
--- a/server/src/lsp-connection.ts
+++ b/server/src/lsp-connection.ts
@@ -12,6 +12,8 @@ import { Logger, LspClientLogger } from './logger';
 import { LspServer } from './lsp-server';
 import { LspClient, LspClientImpl } from './lsp-client';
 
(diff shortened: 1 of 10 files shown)


AlexTugarev:
this can be tested in Theia with a published `0.4.0-dev.74a2836` version:

screen shot 2019-03-08 at 09 49 21

rchl:
Since the spec for type hierarchy is now finalized and this one is not compliant with that, I will close it.

It's interesting though that even VSCode doesn't implement type hierarchy for typescript. I wonder if that's because there doesn't seem to an appropriate API for that in TSServer and the implementation would have to be a bit of a hack.

PR #125, opened 2019-10-06
Fix empty documentHighlight results due to inconsistent path delimiters

### LINKED ISSUE #94: Issue with documentHighlight feature
Hello,

I'm indirectly using this language server through the [Eclipse WildWebDeveloper tool](https://github.com/eclipse/wildwebdeveloper). I encountered an issue with selection highlighting which we think may be caused by the typescript language server itself (more information [here](https://github.com/eclipse/wildwebdeveloper/issues/104)).

Here is the request that is sent to the server:
```
{
   "jsonrpc":"2.0",
   "id":"12",
   "method":"textDocument/hover",
   "params":{
      "textDocument":{
         "uri":"file:///[...]/popup.js"
      },
      "uri":"file:///[...]/popup.js",
      "position":{
         "line":95,
         "character":32
      }
   }
}
```

_popup.js_ is a file checked out locally, but corresponds to the Javascript document located in [this repository](https://github.com/PyvesB/night-video-tuner/blob/master/popup/popup.js). I'm trying to highlight the `storageObj` variable.

However, the response from the server comes back empty:
```
{
   "jsonrpc":"2.0",
   "id":"12",
   "result":{
      "contents":[

      ]
   }
}
```

Is this to be expected?

Thanks in advance for your help.

Cheers,

Pyves

PyvesB:
Closes #94.

I did not provide any additional unit tests as the only realistic setup to reproduce this issue would be when running the test suite on a Windows machine, which is impractical. The change is small and well documented, it's probably not worth it anyway!

### DIFF
--- a/server/src/lsp-server.ts
+++ b/server/src/lsp-server.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017, 2018 TypeFox and others.
+ * Copyright (C) 2017 - 2019 TypeFox and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
@@ -758,7 +758,9 @@ export class LspServer {
         }
         const result: lsp.DocumentHighlight[] = [];
         for (const item of response.body) {
-            if (item.file === file) {
+            // tsp returns item.file with POSIX path delimiters, whereas file is platform specific.
+            // Normalizing both ensures that the comparison is consistent.
+            if (path.normalize(item.file) === path.normalize(file)) {
                 const highlights = toDocumentHighlight(item);
                 result.push(...highlights)
             }
--- a/server/src/lsp-server.ts
+++ b/server/src/lsp-server.ts
@@ -757,10 +757,11 @@ export class LspServer {
             return []
         }
         const result: lsp.DocumentHighlight[] = [];
+        const normalizedFile = path.normalize(file);
         for (const item of response.body) {
             // tsp returns item.file with POSIX path delimiters, whereas file is platform specific.
             // Normalizing both ensures that the comparison is consistent.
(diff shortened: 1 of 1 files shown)


server/src/lsp-server.ts
```diff
@@ -758,7 +758,9 @@ export class LspServer {
         }
         const result: lsp.DocumentHighlight[] = [];
```
princemaple: probably better move the normalization of the `file` up to here, like

```js
const normalizedFilePath = path.normalize(file);
```

princemaple:
@akosyakov Could you please take a look? thanks

server/src/lsp-server.ts
PyvesB: Done, thanks.

princemaple:
Is there an option to request a review from the maintainers? @PyvesB

server/src/lsp-server.ts
```diff
@@ -757,8 +757,11 @@ export class LspServer {
             return []
         }
         const result: lsp.DocumentHighlight[] = [];
+        const normalizedFile = path.normalize(file);
```
akosyakov: should not be necessary, file is already normalized via `URI.parse(uri).fsPath`

server/src/lsp-server.ts
```diff
@@ -757,8 +757,11 @@ export class LspServer {
...
         for (const item of response.body) {
-            if (item.file === file) {
+            // tsp returns item.file with POSIX path delimiters, whereas file is platform specific.
+            // Normalizing both ensures that the comparison is consistent.
+            if (path.normalize(item.file) === normalizedFile) {
```
akosyakov: I would try to use something like `uriToPath(pathToUri(itemfile)) === file`

PyvesB:
Thanks for the feedback @akosyakov! How does the alternative approach look?

princemaple:
@akosyakov any chance this one gets released?

PR #131, opened 2019-12-15
Call compilerOptionsForInferredProjects during initialization

danielmartin:
Fixes https://github.com/eclipse-theia/theia/issues/6720

In order to get better code intelligence for "inferred" projects,
tsserver supports a "compilerOptionsForInferredProjects" command that
informs the TypeScript compiler about how files should be compiled.

This mimicks what the VS Code client does just after calling
"configure" (https://github.com/microsoft/vscode/blob/58d954737facf9478a8255965e5d365bc3d70f71/extensions/typescript-language-features/src/typescriptServiceClient.ts#L479).

Example of better code completion when working on the React sample project (https://code.visualstudio.com/docs/nodejs/reactjs-tutorial):

**Before:**

Screen Shot 2019-12-16 at 12 32 33 AM

**After:**

Screen Shot 2019-12-16 at 12 29 01 AM

### DIFF
--- a/server/src/lsp-server.ts
+++ b/server/src/lsp-server.ts
@@ -128,6 +128,17 @@ export class LspServer {
             }
         });
 
+        this.tspClient.request(CommandTypes.CompilerOptionsForInferredProjects, {
+            options: {
+                module: tsp.ModuleKind.CommonJS,
+                target: tsp.ScriptTarget.ES2016,
+                jsx: tsp.JsxEmit.Preserve,
+                allowJs: true,
+                allowSyntheticDefaultImports: true,
+                allowNonTsExtensions: true
+            }
+        });
+
         const logFileUri = logFile && pathToUri(logFile, undefined);
         this.initializeResult = {
             capabilities: {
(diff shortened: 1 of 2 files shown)


danielmartin:
I planned to add a test for this code change, but I don't quite understand how the completion tests work. According to https://github.com/danielmartin/typescript-language-server/blob/3114c557b58eb6b286e3c63f68ffe56c26800224/server/src/lsp-server.spec.ts#L48-L53, that particular test seems to send the cursor position just before "console" to the language server, so I assume the list of "proposals" is a list of more than 800 globally accessible symbols? If I modify that test to position at `position(doc, 'log');` instead, I would expect a list of symbols from `console`, but that's not what I get. Am I understanding this wrong? I'm quite new to this codebase and to JavaScript and TypeScript development in general.

DonnieWest:
This LGTM - I'll add my own tests separately

I'll also note, currrently the `tsserver` will throw out some erroneous completions, so that that's part of the reason `proposals` is as long as they are. It's bit weird though to stringify those completions and compare the string. That's something that should be fixed

PR #132, opened 2019-12-23
Update command line option tssserver-log-verbosity: off will disable logging 

thanhvg:
`tsserver` can provide options to turn of logging. And for most use cases it is recommended to [turn off  logging.](https://github.com/microsoft/TypeScript/issues/30967)

This PR enables that option.

Thanks.


DonnieWest:
Thanks!

PR #139, opened 2020-02-05
Allow extending LSPServer class

lucasterra:
Hi. I'm writing an LSP Server based on your language server.

We have a product which uses a subset of TypeScript features, and we have our own compiler for it.
The language server I'm writing will extend your TypeScript language server, and add some extra functionality to it. For that, I need to extend the [LspServer class](https://github.com/theia-ide/typescript-language-server/blob/3114c557b58eb6b286e3c63f68ffe56c26800224/server/src/lsp-server.ts#L50-L58). e.g.:

```ts
import { LspServer } from 'typescript-language-server';

class CustomLanguageServer extends LspServer {
    // custom doRequestDiagnostics
    protected async doRequestDiagnostics(): Promise<tsp.RequestCompletedEvent> {
      // call custom compiler
      // call tsserver
    }
}
```

But, in order to call the `tsserver`, I need access to the `tspClient`, which is currently protected by `private` keyword. This PR changes it to `protected`.

-------

I also re-exported the modules in your language server for easy access, so I can write this:
```ts
// write this
import { LspServer } from 'typescript-language-server';
// instead of this
import { LspServer } from "typescript-language-server/lib/lsp-server";
```


DonnieWest:
@lucasterra howdy! Thanks for the PR

Unfortunately, I don't think this is a use case I'd like to support.  We would effectively go from being a developer tool to a library (and all that entails, like not breaking users of the library, keeping stable interfaces, etc) and that's not something I'd like to maintain long term.

Best of luck to your LSP implementation

PR #148, opened 2020-05-28
[WIP] Resolve completions within LspServer.completion()

leungbk:
This allows additionalTextEdits to be provided in the
textDocument/completion response while abiding by the guideline here:

https://github.com/microsoft/language-server-protocol/issues/924

I learned of this today via https://github.com/emacs-lsp/lsp-mode/issues/1713#issuecomment-635513017

### DIFF
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -13,7 +13,8 @@ import { asRange, toTextEdit, asPlainText, asTagsDocumentation, asDocumentation 
 import { Commands } from './commands';
 
 export interface TSCompletionItem extends lsp.CompletionItem {
-    data: tsp.CompletionDetailsRequestArgs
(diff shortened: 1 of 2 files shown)


rchl:
In which cases `additionalTextEdits` are provided for completion response?

It might be a correct fix but not sure if it's worth it as it will make the completion response really really slow (like few seconds instead of instantaneous). I've tried on this server's codebase.

leungbk:
> In which cases `additionalTextEdits` are provided for completion response?

One example is when you're autocompleting a symbol not yet in the current file, and `additionalTextEdits` allows you to auto-import.

> It might be a correct fix but not sure if it's worth it as it will make the completion response really really slow (like few seconds instead of instantaneous). I've tried on this server's codebase.

Hmm, you're right. This PR will need some additional work.

PR #156, opened 2020-10-13
Bump dependencies

DonnieWest:
Bumping some basic dependencies. Running into an issue currently where the bumped `lerna` version appears to randomly fail but not report the build failure

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
         "ts-node": "7.0.1",
         "tslint": "^5.11.0",
         "tslint-language-service": "^0.9.9",
-        "typescript": "^3.0.1"
+        "typescript": "^3.9.0"
     },
     "scripts": {
         "prepare": "lerna run clean && yarn build && yarn bundle",
--- a/server/src/calls.ts
+++ b/server/src/calls.ts
@@ -191,7 +191,7 @@ async function findNonDefinitionReferences(tspClient: TspClient, args: tsp.FileS
     return (await findReferences(tspClient, args)).filter(ref => !ref.isDefinition);
 }
 
-async function findReferences(tspClient: TspClient, args: tsp.FileSpan): Promise<tsp.ReferencesResponseItem[]> {
+async function findReferences(tspClient: TspClient, args: tsp.FileSpan): Promise<readonly tsp.ReferencesResponseItem[]> {
     const file = args.file;
     const result = await tspClient.request(CommandTypes.References, {
         file,
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 lib
 node_modules
 *.log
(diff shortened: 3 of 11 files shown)


server/src/calls.ts
```diff
@@ -191,7 +191,7 @@ async function findNonDefinitionReferences(tspClient: TspClient, args: tsp.FileS
     return (await findReferences(tspClient, args)).filter(ref => !ref.isDefinition);
 }
 
-async function findReferences(tspClient: TspClient, args: tsp.FileSpan): Promise<tsp.ReferencesResponseItem[]> {
+async function findReferences(tspClient: TspClient, args: tsp.FileSpan): Promise<readonly tsp.ReferencesResponseItem[]> {
```
apexskier, line 194: Do you know if there's a difference between these two versions?

```suggestion
async function findReferences(tspClient: TspClient, args: tsp.FileSpan): Promise<ReadonlyArray<tsp.ReferencesResponseItem>> {
```

apexskier (approved):
I'd recommend running `npx yarn-deduplicate yarn.lock && yarn` to deduplicate the lockfile.

apexskier:
I'd also add `tsconfig.buildinfo` to the gitignore - I see that file generated when I check this out locally.

package.json
```diff
@@ -14,7 +14,7 @@
         "ts-node": "7.0.1",
         "tslint": "^5.11.0",
         "tslint-language-service": "^0.9.9",
-        "typescript": "^3.0.1"
+        "typescript": "^3.9.0"
     },
     "scripts": {
         "prepare": "lerna run clean && yarn build && yarn bundle",
```
apexskier, line 20: Actually, I see an issue here locally - build runs lint and compile in parallel, but the example file depends on a compiled file in the server.

```
$ tslint -c ../tslint.json --project .
[compile] example/src/client.ts(13,41): error TS2307: Cannot find module 'typescript-language-server/lib/ts-protocol' or its corresponding type declarations.
[compile] example/src/client.ts(61,36): error TS2345: Argument of type 'null' is not assignable to parameter of type 'IPosition'.
[compile] example/src/client.ts(61,58): error TS2339: Property 'position' does not exist on type 'CancellationToken'.
```

apexskier: Looks like this was just because I had old/invalid .tsbuildinfo files

apexskier: https://github.com/theia-ide/typescript-language-server/pull/169

server/src/calls.ts
DonnieWest: Looks like there's not much of a practical difference as of TS 3.4 https://stackoverflow.com/questions/48300869/difference-between-readonly-and-readonlyarray#48301181

PR #168, opened 2020-11-13
Ensure TSP request cancellation cancels timeout handling

apexskier:
I'm trying to investigate https://github.com/apexskier/nova-typescript/issues/91, and discovered a potential issue that this fixes. If a typescript server request times out _after_ being cancelled, I think the timeout can still cause an unhandled rejection. I haven't found a real reproduction, so this is purely theoretical.

This fixes that by adding a `cancel` method to the `Deferred` utility that clears the timeout timer.

I also added a few type-only improvements (removing an unused export and some `any`s).

This conflicts a bit with https://github.com/theia-ide/typescript-language-server/pull/92, and can be closed if that one is merged (closed by #92).

### DIFF
--- a/server/src/tsp-client.ts
+++ b/server/src/tsp-client.ts
@@ -137,9 +137,12 @@ export class TspClient {
     ): Promise<TypeScriptRequestTypes[K][1]> {
         this.sendMessage(command, false, args);
         const seq = this.seq;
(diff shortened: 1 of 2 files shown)


DonnieWest (approved):
Thanks! Seems like a reasonable change

PR #170, opened 2020-11-13
Refactor code actions to better support filtering against "only"

apexskier:
This refactors code action processing with the primary goal of exposing an organize imports code action (https://github.com/apexskier/nova-typescript/issues/104).

Currently, this action is only provided if the LSP client requests _only_ organize imports code actions, which I interpret as not quite being to spec. The only argument is documented as such:

> Requested kind of actions to return.
>
> Actions not of this kind are filtered out by the client before being shown, so servers can omit computing them.

I refactored so the `typescript-language-server` makes a request to `tsserver` and fully relies on its [built in functionality](https://devblogs.microsoft.com/typescript/announcing-typescript-2-8-2/#organize-imports). This appears to be slightly different than how VSCode works, which doesn't show this action in the code actions pane, but I think will be more intuitive, since clients won't have to expose a separate UI/command to fetch this code action.

I refactored the `only` processing to more correctly optimize code action requests - this will now fully skip the request to the tsserver if only is present and the type of code action is not provided. I also refactored to use generators when processing `tsserver` responses instead of passing an array in and directly mutating it, which makes these functions easier to test and work with, in my opinion.

I also added a server-level test for the code actions functionality and a unit test for the organize import processing.


rchl:
> Currently, this action is only provided if the LSP client requests _only_ organize imports code actions

I can't confirm that that's the case with the current code.

See this exchange:


```
-->

{
  "context": {
    "diagnostics": []
  },
  "range": {
    "end": {
      "character": 7,
      "line": 10
    },
    "start": {
      "character": 7,
      "line": 10
    }
  },
  "textDocument": {
    "uri": "file:///foo.js"
  }
}

<--

[
  {
    "command": {
      "arguments": [
        "/foo.js"
      ],
      "command": "_typescript.organizeImports",
      "title": ""
    },
    "kind": "source.organizeImports",
    "title": "Organize imports"
  }
]
```

This is problematic because, for editors that show an indication that there is code action available, those will show such an indication for every line in the document.

I'm fine with you returning this code action when `only` is explicitly set. This is in fact the convention that is already in place in many clients and used to request those code actions on saving. For example in VSCode you'd set:
```js
  "editor.codeActionsOnSave": {
    "source.organizeImports": true
  },
```
which would result in an editor requesting `source.organizeImports` code actions from servers that support this code action and automatically applying edits on saving.

But! You must also ensure that the server announces support for that code action in `codeActionProvider.codeActionKinds` which you are also missing in your change.

server/src/organize-imports.ts
```diff
@@ -8,12 +8,20 @@
 import * as lsp from 'vscode-languageserver';
 import * as tsp from 'typescript/lib/protocol';
 import { Commands } from './commands';
+import { CodeActionKind } from "vscode-languageserver";
 
-export function provideOrganizeImports(file: string, context: lsp.CodeActionContext, result: (lsp.Command | lsp.CodeAction)[]): void {
-    if (!context.only || context.only.indexOf(lsp.CodeActionKind.SourceOrganizeImports) === -1) {
+export function *provideOrganizeImports(
```
rchl, line 13: Why have you switched to generators? Is there any benefit in that?

I think it would only make sense if the calling code does something with the result before the full result is obtained. In the case of getting diagnostics, the client needs the full list anyway so a simple loop seems like will be more efficient than generators which are essentially separate function calls to get each and every item.

apexskier: > I also refactored to use generators when processing tsserver responses instead of passing an array in and directly mutating it, which makes these functions easier to test and work with, in my opinion.

I can return a generated array, I mostly didn't like the passing of a mutable array in pattern

apexskier:
> This is problematic because, for editors that show an indication that there is code action available, those will show such an indication for every line in the document.

> I refactored so the typescript-language-server makes a request to tsserver and fully relies on its built in functionality. This appears to be slightly different than how VSCode works, which doesn't show this action in the code actions pane, but I think will be more intuitive, since clients won't have to expose a separate UI/command to fetch this code action.

I knew this would be the biggest concern with this change, but I can revert to the current behavior.

apexskier:
> But! You must also ensure that the server announces support for that code action in codeActionProvider.codeActionKinds server capability, which you are also missing in your change.

I'm not positive, but I _think_ I don't need to change anything here now that the external behavior is the same as it previously was.

rchl:
> > But! You must also ensure that the server announces support for that code action in codeActionProvider.codeActionKinds server capability, which you are also missing in your change.
> 
> I'm not positive, but I _think_ I don't need to change anything here now that the external behavior is the same as it previously was.

You can argue that it's out of the scope of this change but then it should be added anyway later.

Without that change, some clients won't take this server into account when triggering "code actions on save".

Also, I believe VSCode will list that code action in the `Source actions...` context menu if it's reported in server capabilities.

server/src/lsp-server.ts
```diff
@@ -38,6 +38,7 @@ import { provideOrganizeImports } from './organize-imports';
 import { TypeScriptInitializeParams, TypeScriptInitializationOptions, TypeScriptInitializeResult } from './ts-protocol';
 import { collectDocumentSymbols, collectSymbolInformations } from './document-symbol';
 import { computeCallers, computeCallees } from './calls';
+import { CodeActionKind } from "vscode-languageserver";
```
rchl, line 41: This code base seems to be using single quotes.

(Unfortunately the used linter doesn't check those...)

server/src/lsp-server.ts
```diff
@@ -673,19 +674,30 @@ export class LspServer {
...
+        if (!params.context.only || params.context.only.findIndex((k: string) => k.startsWith(CodeActionKind.QuickFix)) > -1) {
+            const errorCodes = params.context.diagnostics.map(diagnostic => Number(diagnostic.code));
+            actions.push(...provideQuickFix(await this.getCodeFixes({ ...args, errorCodes }), this.documents));
+        }
+        if (!params.context.only || params.context.only.findIndex((k: string) => k.startsWith(CodeActionKind.Refactor)) > -1) {
+            actions.push(...provideRefactors(await this.getRefactors(args), args));
+        }
+
+        // organize import is provided by tsserver for any line, so we only get it if explicitly requested
+        if (params.context.only && params.context.only.indexOf(CodeActionKind.SourceOrganizeImports) > -1) {
```
rchl, line 694: The logic for checking `only` should be the same here as in other cases - should use `startsWith`. Otherwise, the code action won't be provided if the client asks for `only: ['source']`.

server/src/organize-imports.ts
```diff
@@ -5,15 +5,20 @@
...
+): Array<lsp.CodeAction> {
+    if (!response) {
+        return [];
     }
-    result.push(lsp.CodeAction.create("Organize Imports",
-        lsp.Command.create('', Commands.ORGANIZE_IMPORTS, file),
-        lsp.CodeActionKind.SourceOrganizeImports));
+    return response.body.map(edit => lsp.CodeAction.create(
+        "Organize imports",
+        lsp.Command.create("", Commands.ORGANIZE_IMPORTS, edit.fileName),
```
rchl, line 21: Is it possible to provide an edit right away instead of the command?

Command requires a round-trip to the server to execute and with the `only` being used to request the code action, there is a very high chance that the user will trigger that code actions.

apexskier: Possible, but I'd prefer to not do this.

We'd need to convert the [typescript server location](https://github.com/microsoft/TypeScript/blob/cec2fda9a53620dc545a2c4d7b0156446ab145b4/lib/protocol.d.ts#L1510-L1531) (`{start: number, end: number}`) to the [LPS location](https://github.com/microsoft/vscode-languageserver-node/blob/d58c00bbf8837b9fd0144924db5e7b1c543d839e/types/src/main.ts#L112-L122), which we need the [full document text to do](https://github.com/theia-ide/typescript-language-server/blob/f43aff23b9eaaae83d11a422fb509667951b2104/server/src/document.ts#L34-L36). At this context, we don't have that and it's not guaranteed that the file has been opened, technically (I suppose someone could request organize imports on an unopened file, since the document URI is the argument).

It'd be possible to try to get the document at the caller of this and pass it optionally into this `provideOrganizeImports` function, but it feels like that'll be more likely to be flaky to me. (a) the main `response` arg is fetched asynchronously and (b) the edits are based on tsserver state and the optional document text I'd pass in would be typescript-language-server state.

apexskier (commented):
Sorry I've been away from this for so long!

server/src/calls.ts
```diff
@@ -1,5 +1,5 @@
 
-import tsp from 'typescript/lib/protocol';
+import * as tsp from 'typescript/lib/protocol';
```
apexskier, line 2: This was showing up as an error in my editor.

server/src/organize-imports.ts
rchl: OK. I think it would be a good improvement to have when the document is open at least but it can be considered later.

server/src/calls.ts
rchl: Maybe because you haven't run `yarn` after rebasing? It shouldn't be needed AFAIK.

apexskier: I think vscode hadn't picked up the tsconfig change. `"esModuleInterop": true` means it should work.

server/src/lsp-server.ts
```diff
@@ -674,19 +675,30 @@ export class LspServer {
...
         }
         const args = toFileRangeRequestArgs(file, params.range);
-        const codeActions: (lsp.Command | lsp.CodeAction)[] = [];
-        const errorCodes = params.context.diagnostics.map(diagnostic => Number(diagnostic.code));
-        provideQuickFix(await this.getCodeFixes({ ...args, errorCodes }), codeActions, this.documents);
-        provideRefactors(await this.getRefactors(args), codeActions, args);
-        provideOrganizeImports(file, params.context, codeActions);
-        return codeActions;
+        const actions: lsp.CodeAction[] = [];
+        if (!params.context.only || params.context.only.findIndex((k: string) => k.startsWith(CodeActionKind.QuickFix)) > -1) {
```
rchl, line 686: Just a style thing but you can use `some` instead of `findIndex`?

server/src/lsp-server.ts
```diff
@@ -674,19 +675,30 @@ export class LspServer {
...
+            const errorCodes = params.context.diagnostics.map(diagnostic => Number(diagnostic.code));
+            actions.push(...provideQuickFix(await this.getCodeFixes({ ...args, errorCodes }), this.documents));
+        }
+        if (!params.context.only || params.context.only.findIndex((k: string) => k.startsWith(CodeActionKind.Refactor)) > -1) {
+            actions.push(...provideRefactors(await this.getRefactors(args), args));
+        }
+
+        // organize import is provided by tsserver for any line, so we only get it if explicitly requested
+        if (params.context.only && params.context.only.findIndex((k: string) => k.startsWith(CodeActionKind.SourceOrganizeImports)) > -1) {
```
rchl, line 695: It seems like the "startsWith" is used "backwards".  For example, the user-provided `only` token with an example value `source.organizeImports.pyright` should not result in this server providing the action since the user-requested one is more specific than `source.organizeImports` (`CodeActionKind.SourceOrganizeImports`).

But also `startsWith` is not a great solution either way because we want to match against full tokens. So `source.fixAll` should match `source.fixAll.eslint` but not `source.fixAllFoo`. So ideally we would split the token at dot and compare the full tokens.

server/src/organize-imports.ts
```diff
@@ -6,14 +6,19 @@
  */
 
 import * as lsp from 'vscode-languageserver';
+import * as tsp from 'typescript/lib/protocol';
```
rchl, line 9:
```suggestion
import tsp from 'typescript/lib/protocol';
```

server/src/organize-imports.ts
```diff
@@ -6,14 +6,19 @@
...
 import { Commands } from './commands';
+import { CodeActionKind } from 'vscode-languageserver';
 
-export function provideOrganizeImports(file: string, context: lsp.CodeActionContext, result: (lsp.Command | lsp.CodeAction)[]): void {
-    if (!context.only || context.only.indexOf(lsp.CodeActionKind.SourceOrganizeImports) === -1) {
-        return;
+export function provideOrganizeImports(
+    response: tsp.OrganizeImportsResponse | undefined
+): Array<lsp.CodeAction> {
```
rchl, line 15: Would be good to keep it one line.

server/src/quickfix.ts
```diff
@@ -6,22 +6,26 @@
  */
 
 import * as lsp from 'vscode-languageserver';
-import tsp from 'typescript/lib/protocol';
+import * as tsp from 'typescript/lib/protocol';
```
rchl, line 9:
```suggestion
import tsp from 'typescript/lib/protocol';
```

server/src/quickfix.ts
```diff
@@ -6,22 +6,26 @@
...
 import { Commands } from './commands';
 import { toTextDocumentEdit } from './protocol-translation';
 import { LspDocuments } from './document';
+import { CodeActionKind } from 'vscode-languageserver';
 
-export function provideQuickFix(response: tsp.GetCodeFixesResponse | undefined, result: (lsp.Command | lsp.CodeAction)[], documents: LspDocuments | undefined): void {
+export function provideQuickFix(
+    response: tsp.GetCodeFixesResponse | undefined,
+    documents: LspDocuments | undefined
+): Array<lsp.CodeAction> {
```
rchl, line 18: Please keep it on one line.

server/src/refactor.ts
```diff
@@ -6,22 +6,26 @@
  */
 
 import * as lsp from 'vscode-languageserver';
-import tsp from 'typescript/lib/protocol';
+import * as tsp from 'typescript/lib/protocol';
```
rchl, line 9:
```suggestion
import tsp from 'typescript/lib/protocol';
```

server/src/refactor.ts
```diff
@@ -6,22 +6,26 @@
...
 import { Commands } from './commands';
 
-export function provideRefactors(response: tsp.GetApplicableRefactorsResponse | undefined, result: (lsp.Command | lsp.CodeAction)[], args: tsp.FileRangeRequestArgs): void {
+export function provideRefactors(
+    response: tsp.GetApplicableRefactorsResponse | undefined,
+    args: tsp.FileRangeRequestArgs
+): Array<lsp.CodeAction> {
```
rchl, line 15: Please keep on one line.

server/src/refactor.ts
```diff
@@ -6,22 +6,26 @@
...
-            for (const action of info.actions) {
-                result.push(asApplyRefactoring(action, info, args));
-            }
+            actions.push(...info.actions.map(action =>
+                asApplyRefactoring(action, info, args)
+            ));
         }
-    }
+        return actions;
+    }, []);
```
rchl, line 28: Isn't this changed version equal to the previous version while being also less readable (IMO)?
Please revert if so because it was fine as is.

rchl (changes requested):

server/src/refactor.ts
apexskier: The main difference is tho return an array, rather than mutate a param. I can convert it to not use `reduce` though.

server/src/lsp-server.ts
apexskier: Can you clarify a bit?

> `source.organizeImports.pyright` should not result in this server providing the action since the user-requested one is more specific than `source.organizeImports`
> 
> `source.fixAll` should match `source.fixAll.eslint` but not `source.fixAllFoo`

I think you're referring to the different between organize imports and other actions (that we only show organize imports if explicitly requested) but I'm not sure. I'll push up what I'm interpreting this as.

server/src/lsp-server.spec.ts
```diff
@@ -456,6 +457,292 @@ describe('signatureHelp', () => {
...
+                                                },
+                                                start: {
+                                                    character: 24,
+                                                    line: 1
+                                                }
+                                            }
+                                        }
+                                    ],
+                                    textDocument: {
+                                        uri: 'file://ROOT/server/test-data/bar.ts',
```
rchl, line 571: Use `uri('bar.ts')` where you expect URI and `filePath('bar.ts')` where you expect file path and remove those hacky replaces.

rchl (changes requested):

server/src/lsp-server.spec.ts
```diff
@@ -645,6 +698,49 @@ describe('code actions', () => {
...
+            {
+                command: {
+                    arguments: ['ROOT/server/test-data/bar.ts'],
+                    command: '_typescript.organizeImports',
+                    title: ''
+                },
+                kind: 'source.organizeImports',
+                title: 'Organize imports'
+            }
+        ]);
```
rchl, line 742: Requesting `source.organizeImports.test` should not result in code action being provided. The more specific `only` should not return the less specific `kind`.

server/src/lsp-server.ts
rchl: Does this comment clarify it https://github.com/theia-ide/typescript-language-server/pull/170#discussion_r661792138 ?

I've touched on two different issues here. The latter is fixed, the former is what I've linked to now.

server/src/lsp-server.spec.ts
apexskier: great!

server/src/lsp-server.ts
apexskier: Thanks, that clarifies!

server/src/lsp-server.spec.ts
apexskier: Got it - I've added more tests to more explicitly cover this.

rchl:
Oh, crap...

I've realized that I've completely mislead you with my comment about the matching logic.

Spec says:

```
	/**
	 * Requested kind of actions to return.
	 *
	 * Actions not of this kind are filtered out by the client before being
	 * shown. So servers can omit computing them.
	 */
	only?: CodeActionKind[];
```

The server should not have any of that `startsWith` or "match partial code action tokens" logic. It's the job of the client to get the code actions supported by the server (the `codeActionProvider.codeActionKinds` capability) and match the user-specified code action tokens against those and actually request the code actions using the tokens that the server reported. The server's job is to match those tokens exactly.

So lets say that the server's `codeActionProvider.codeActionKinds` capability reports `source.organizeImports.typescript` being supported. If the user has specified in the client that `source.organizeImports` actions should run on save then the client does the matching and actually sets `only` to `source.organizeImports.typescript`.

I'm partially responsible for you making all those extra changes although your initial changes also used `startsWith` so maybe that influenced my thinking how this should work.

Of course you don't have to make changes to `codeActionProvider.codeActionKinds` capability as that would change some behavior (and I also found a bug with that when using code actions on save) but you'd have to change to matching to just match whole tokens.

apexskier:
Well that simplifies things, at least. We can just use an `includes` check.

rchl (approved):
Looks great now :)

rchl:
But looking at all the changes we ended up with here, I can't quite see any actual behavior change after those changes. Looks more like a refactor now (nice refactor but still just that).

rchl:
Thanks!

apexskier:
Yeah, one of the reasons I semi-abandoned this is because you called out that it doesn't make a ton of sense to show the organize imports action on every line (https://github.com/theia-ide/typescript-language-server/pull/170#issuecomment-727010796). I ended up adding a manually run command in my extension instead of relying on this.

I think the only real runtime impact is:
> I refactored the only processing to more correctly optimize code action requests - this will now fully skip the request to the tsserver if only is present and the type of code action is not provided.

I decided to finish this out because I want to make some additional changes to support https://github.com/apexskier/nova-typescript/issues/273 and didn't want to leave it unfinished.

rchl:
> Yeah, one of the reasons I semi-abandoned this is because you called out that it doesn't make a ton of sense to show the organize imports action on every line ([#170 (comment)](https://github.com/theia-ide/typescript-language-server/pull/170#issuecomment-727010796)). I ended up adding a manually run command in my extension instead of relying on this.

Yes, and I stand by that still.

I think the next step (not necessarily for you) would be to expose that action kind in `codeActionProvider.codeActionKinds` capability. That will result in it being exposed in the `Source actions...` context menu (at least in VSCode and Sublime Text) and also allow to use it within the "code actions on save" feature. There is currently an issue with the latter though. It triggers infinite loop (that is cut short after a second or so by the editor) because typescript reports edits even if nothing is actually modified. Might get to reporting it in typescript after looking at it a bit more.

apexskier:
> > Yeah, one of the reasons I semi-abandoned this is because you called out that it doesn't make a ton of sense to show the organize imports action on every line ([#170 (comment)](https://github.com/theia-ide/typescript-language-server/pull/170#issuecomment-727010796)). I ended up adding a manually run command in my extension instead of relying on this.
> 
> Yes, and I stand by that still.

(FYI, I agree with you here, didn't mean to sound like I didn't)

PR #172, opened 2020-11-18
[WIPfeat(diagnostics): ability to ignore some codes

pyrho:


server/src/cli.ts
```diff
@@ -45,5 +46,6 @@ createLspConnection({
     tsserverPath: program.tsserverPath as string,
     tsserverLogFile: program.tsserverLogFile as string,
     tsserverLogVerbosity: program.tsserverLogVerbosity as string,
-    showMessageLevel: logLevel as lsp.MessageType
+    showMessageLevel: logLevel as lsp.MessageType,
+    ignoredDiagnosticCodes: program.ignoredDiagnosticCodes || [] as number[],
```
rchl, line 50: `[] as number[]` => `number[]`?

server/src/diagnostic-queue.ts
```diff
@@ -49,18 +49,23 @@ export class DiagnosticEventQueue {
...
     updateDiagnostics(kind: EventTypes, event: tsp.DiagnosticEvent): void {
         if (!event.body) {
             this.logger.error(`Received empty ${event.event} diagnostics.`)
             return;
         }
-        const { file } = event.body;
+        const { file, diagnostics: diagnosticsFromServer } = event.body;
+
+        // `d.code` can be undefined
+        const isDiagnosticAllowed = (d: tsp.Diagnostic) => !d.code || this.ignoredDiagnosticCodes.indexOf(d.code) === -1;
```
rchl, line 64: `[].includes()` instead of `indexOf` maybe?

server/src/lsp-connection.ts
```diff
@@ -28,15 +35,18 @@ export function createLspConnection(options: IServerOptions): lsp.IConnection {
...
 
     connection.onInitialize(server.initialize.bind(server));
 
     connection.onDidOpenTextDocument(server.didOpenTextDocument.bind(server));
     connection.onDidSaveTextDocument(server.didSaveTextDocument.bind(server));
     connection.onDidCloseTextDocument(server.didCloseTextDocument.bind(server));
-    connection.onDidChangeTextDocument(server.didChangeTextDocument.bind(server));
+    connection.onDidChangeTextDocument(
+        server.didChangeTextDocument.bind(server)
+    );
```
rchl, line 49: Was this formatting change necessary? (Also below this one)

pyrho: Indeed, a mistake I was wrestling with my Prettier plugin :P
Will amend.

server/src/lsp-server.spec.ts
```diff
@@ -184,6 +184,31 @@ describe('diagnostics', () => {
...
+            textDocument: doc
+        })
+
+        server.requestDiagnostics();
+        await server.requestDiagnostics();
+        await new Promise(resolve => setTimeout(resolve, 200));
+        const diagnosticsForThisFile = diagnostics.filter(d => d!.uri === doc.uri);
+        assert.equal(diagnosticsForThisFile.length, 1, JSON.stringify(diagnostics));
+        const fileDiagnostics = diagnosticsForThisFile[0]!.diagnostics;
+        assert.equal(fileDiagnostics.length, 0);
```
pyrho, line 210: This test passes but only because the sample TS code does not generate a diagnostic.

server/src/lsp-connection.ts
pyrho: Did [amend](https://github.com/pyrho/typescript-language-server/blob/feat/ignore-diagnostics/server/src/lsp-connection.ts) this change, but this PR isn't updated yet, not sure if it's github lagging or if I'm doing something wrong 😅

server/src/diagnostic-queue.ts
rchl: Array.includes is supported since node 6 though so it won't be a problem.

pyrho: This project [currently targets](https://github.com/theia-ide/typescript-language-server/blob/master/tsconfig.settings.json#L7) es6 (or 2015), and it [looks like](https://github.com/microsoft/TypeScript/blob/master/src/lib/es2016.array.include.d.ts) `Array.includes` is only available for es7 (2016) compilation targets.

server/src/lsp-server.spec.ts
pyrho: But in reality, a diagnostic is emitted 
Screenshot 2020-11-19 at 12 03 35

server/src/diagnostic-queue.ts
rchl: I see. Then never mind me although that target should probably be updated in a separate PR. That target is very old.

server/src/lsp-connection.ts
rchl: It looks good now. The change doesn't appear in the diff anymore.

server/src/diagnostic-queue.ts
pyrho: I guess targeting an older standard has the added benefit of maximizing compatibility, if for some reason you're stuck with an older node version.

rchl: But ES7 is supported from Node7 (or Node 8 if you prefer "stable" release. It would be reasonable to require at least Node 8 as that is already 3 or 4 years old.

pyrho: Agreed that targeting the current [LTS version](https://nodejs.org/en/about/releases/) would be better.

DonnieWest: I'd lean toward getting the target bumped prior to this PR if either of you are interested. Based on https://nodejs.org/en/about/releases/ we should probably support Node v10 and up

server/src/lsp-server.spec.ts
DonnieWest: Hrm, I'm not getting the same diagnostic error. I wonder what's up

But, switching the code to 
```
x() {
    return Promise.resolve(1).then(a => a + 1);
}
```

Does give me a diagnostic error:

```
  fileDiagnostics: [
    {
      range: [Object],
      message: "Cannot find name 'x'.",
      severity: 1,
      code: 2304,
      source: 'typescript',
      relatedInformation: undefined
    }
  ]
}

```

Which I can ignore by setting `ignoredDiagnosticCodes: [2304],` and that makes the tests pass

pyrho: I'll edit the test for this then.

pyrho: Maybe it has to do with the typescript version in use, or the target; I'm unclear as to which settings are used in the tests.

pyrho: Pushed the test change on [this fixup commit](https://github.com/pyrho/typescript-language-server/commit/c1666a2044951ea74b40739a274ae85dcaae932c)

server/src/diagnostic-queue.ts
pyrho: I agree bumping deps and updating the target JS version is high priority, but it's unrelated to the changes in this PR and shouldn't be a blocker for merging this imo.

prabirshrestha:
seems like should pass as workspace config instead of commandline. or could support both

rchl:
Good point IMO. And I would not introduce a command-line option for something like that. Command-line options should be reserved for toggling things that can't be changed once the server is started or are global in nature that can't be controlled with workspace settings.

pyrho:
> seems like should pass as workspace config instead of commandline. or could support both

Agreed that this would be better, does typescript-language-server already support such configuration option (via an external config file) ? If not I think this is outside the scope of this PR.
If yes, could you point me in the right direction?

Thanks.

rchl:
> > seems like should pass as workspace config instead of commandline. or could support both
> 
> Agreed that this would be better, does typescript-language-server already support such configuration option (via an external config file) ? If not I think this is outside the scope of this PR.
> If yes, could you point me in the right direction?

The client can provide setting through the `params` object to the [`initialize` request](https://microsoft.github.io/language-server-protocol/specification#initialize).
It's handled here:
https://github.com/theia-ide/typescript-language-server/blob/fdf28313833cd6216d00eb4e04dc7f00f4c04f09/server/src/lsp-server.ts#L100-L104

rchl:
(there is also [workspace/configuration](https://microsoft.github.io/language-server-protocol/specification#workspace_configuration) which is more dynamic and can be set in more granular way in IDEs but since that's not currently supported, it would be out of scope to add it.

pyrho:
Thank you @rchl, according to neovim's LSP [doc](https://neovim.io/doc/user/lsp.html#vim.lsp.start_client()) this is doable, I will look into it when I get a chance.

prabirshrestha:
vim-lsp supports it too. You can change it via vimrc.

PR #189, opened 2021-03-03
Remove hardcoded request timeouts

harto:
Removes timeouts from `Deferred`, because clients can choose to cancel long-running operations if they need to.

This is the same as #92 with merge conflicts resolved.


DonnieWest:
Thanks!

PR #195, opened 2021-03-27
Use `require.resolve` for module resolution

### LINKED ISSUE #193: PNP Support?
I'm not sure if my setup is the problem or if yarn plug n' play isn't supported. I can't seem to get module's to resolve in a yarn 2 environment. I tried using the yarn sdk's (https://yarnpkg.com/getting-started/editor-sdks#emacs) aswell to see if It was that but it still is unable to resolve any of my installed dependencies. The project runs fine; I'm using the default vue w/ typescript from vite. I can provide more information but I wanted to see if maybe it just wasnt supported yet first.

Jaden-Giordano:
Rather than using a hardcoded search through the node_modules directory use `require.resolve`. This allows for typescript-language-server to be compatible with Yarn 2 projects. This aims to resolve #193


server/src/modules-resolver.ts
```diff
@@ -11,7 +11,7 @@ import * as paths from 'path';
 export function findPathToModule(dir: string, moduleName: string): string|undefined {
     const stat = fs.statSync(dir)
     if (stat.isDirectory()) {
-        const candidate = paths.resolve(dir, 'node_modules', moduleName)
+        const candidate = require.resolve(moduleName, { paths: [dir] });
         if (fs.existsSync(candidate)) {
```
merceyz, line 15: `require.resolve` will throw if it can't find it so this check isn't needed, however this entire function can be replaced with 
```js
try {
  return require.resolve(moduleName, { paths: [dir] });
} catch {
  return undefined;
}
```

server/src/modules-resolver.ts
```diff
@@ -5,20 +5,10 @@
...
-        const candidate = paths.resolve(dir, 'node_modules', moduleName)
-        if (fs.existsSync(candidate)) {
-            return candidate;
-        }
-    }
-    const parent = paths.resolve(dir, '..')
-    if (parent !== dir) {
-        return findPathToModule(parent, moduleName)
+    try {
+        return require.resolve(moduleName, { paths: [dir] });
```
merceyz, line 10:
```suggestion
        return fs.dirname(require.resolve(`${moduleName}/package.json`, { paths: [dir] }));
```
Seems like it actually wants the folder and not the main file

EliaECoyote:
Disclaimer: I don't know much about LSP and theia.

I tested this change in a private yarn-based repo, and I noticed that it's not picking up the `.yarn/sdks/typescript/bin/tsserver` executable correctly (generated through `yarn dlx @yarnpkg/pnpify --sdk vim` as described in the [Yarn editor-sdks docs](https://yarnpkg.com/getting-started/editor-sdks#vim--cocnvim)).

I wonder if this PR should also update `findPathToModule` to use something similar to `const yarnSdkCandidate = paths.resolve(dir, '.yarn/sdks', moduleName)` fn in order to overcome this.

What do you think about it?

js-sketch:
Will this also support opening Yarn's cache archives?

This would make working with zero-install repos viable: https://yarnpkg.com/features/zero-installs

server/src/modules-resolver.ts
rchl: Actually no. The naming of the argument `moduleName` is confusing because it's not really a module name. The two call sites that use it pass something like:
 - `.bin/tsserver`
 - `typescript/lib/tsserver.js`
And the returned value is expected to be a full path to those scripts.

rchl (approved):
LGTM

DonnieWest:
@rchl doesn't #220 supecede this one?

rchl:
> @rchl doesn't #220 supecede this one?

I think this is how it was supposed to be from the beginning and it will fix some cases.

#220 goes to another level and handles some very specific use cases that I'm not very comfortable with so we could let that other one brew for a bit more.

PR #207, opened 2021-04-29
Add ability to pass down userPreferences to the LSP through the CLI

rmagatti:
Draft work for adding the ability to pass down `UserPreferences` to the LSP.

### DIFF
--- a/server/src/lsp-server.ts
+++ b/server/src/lsp-server.ts
@@ -93,7 +93,8 @@ export class LspServer {
         return bundled;
     }
 
-    async initialize(params: TypeScriptInitializeParams): Promise<TypeScriptInitializeResult> {
+    initialize(preferences?: tsp.UserPreferences) {
+      return async function(params: TypeScriptInitializeParams): Promise<TypeScriptInitializeResult> {
         this.logger.log('initialize', params);
         this.initializeParams = params;
 
@@ -124,7 +125,8 @@ export class LspServer {
         this.tspClient.start();
         this.tspClient.request(CommandTypes.Configure, {
             preferences: {
-                allowTextChangesInNewFiles: true
+                allowTextChangesInNewFiles: true,
+                ...preferences
             }
         });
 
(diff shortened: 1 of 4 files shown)


server/src/lsp-server.ts
```diff
@@ -93,91 +93,94 @@ export class LspServer {
...
-            logger: this.options.logger,
-            onEvent: this.onTsEvent.bind(this)
-        });
-
-        this.tspClient.start();
-        this.tspClient.request(CommandTypes.Configure, {
-            preferences: {
-                allowTextChangesInNewFiles: true
+    initialize(preferences?: tsp.UserPreferences) {
+        return async function(params: TypeScriptInitializeParams): Promise<TypeScriptInitializeResult> {
```
rmagatti, line 97: There's probably a better way to do this, but for now I just curried the initialize function to pass down the user preferences.

server/src/lsp-server.ts
```diff
@@ -93,91 +93,94 @@ export class LspServer {
...
-                hoverProvider: true,
-                renameProvider: true,
-                referencesProvider: true,
-                signatureHelpProvider: {
-                    triggerCharacters: ['(', ',', '<']
+            this.tspClient.start();
+            this.tspClient.request(CommandTypes.Configure, {
+                preferences: {
+                    allowTextChangesInNewFiles: true,
+                    ...preferences
```
rmagatti, line 129: Used here.

Every other change in this file is merely identation.

rchl:
Why are those options not passed through the `initialize` request and instead on the command line? Is there something preventing it?

rmagatti:
> Why are those options not passed through the `initialize` request and instead on the command line? Is there something preventing it?

Hm, probably just my lack of understanding?
First time even looking at this code really, just needed to make this work for my own local workflow, thought I'd put at least a draft up to get precisely this kind of feedback. I admittedly do not understand this codebase as much as I'd like to yet :)

I can dig into it a bit more and change the approach if something else makes more sense though 😊

rmagatti:
#218 has a better implementation than this

PR #220, opened 2021-06-14
Support Yarn PnP

tapayne88:
Building on what's been discussed in #195 

Avoided converting to `require.resolve` as yarn v2 uses a specific location for editor sdk which is created when you run `yarn dlx @yarnpkg/pnpify --sdk vim`.


tapayne88:
While this does get the typescript language server up and running for yarn & yarn v2 projects, vim seems to still need help for things like jump to definition.

This helps in some instances but not for virtualised packages (they include `$$virtual` in the filename and don't seem to exist on disk) - haven't figured out a solution for those.
```vim
function! OpenZippedFile(f)
  " get number of new (empty) buffer
  let l:b = bufnr('%')
  " construct full path
  let l:f = 'zipfile:' . getcwd() . '/' . substitute(a:f, '.zip/', '.zip::', '')

  " swap back to original buffer
  b #
  " delete new one
  exe 'bd! ' . l:b
  " open buffer with correct path
  sil exe 'e ' . l:f
  " read in zip data
  call zip#Read(l:f, 1)
endfunction

augroup yarngtd
  autocmd!

  autocmd BufReadCmd */.yarn/*.zip/* call OpenZippedFile(expand('<afile>'))
augroup END
```

DonnieWest:
@tapayne88 I haven't really used PnP or Yarn v2, is there a chance I could get a sample repository to test this against?

As far as virtualized files, I'll poke around on how others are handling this. I know kotlin-language-server has it's own (non-standard?) way of handling those and we might look to them for inspiration

tapayne88:
@DonnieWest I've created a create-react-app repo using TS and PnP
https://github.com/tapayne88/yarn-pnp-demo

It should run following
```
yarn install
yarn start
```

With `src/App.test.tsx` open, I find using this PR's changes that jump-to-definition (with the above snippet) works on React import but fails on the @testing-library/react import due to virtual location
`/Users/me/git/yarn-pnp-demo/.yarn/$$virtual/@testing-library-react-virtual-3f9689214d/0/cache/@testing-library-react-npm-11.2.7-3a0469c756-389c9f3e83.zip/node_modules/@testing-library/react/types/index.d.ts`

Also find `$$` seems to expand to what I _think_ is the process ID

tapayne88:
Also something to note is I haven't been able to verify the current changes against a windows setup

merceyz:
https://github.com/yarnpkg/berry/pull/2598/files#diff-2dde4b1afd9df91455f77a86893b55861d1be4e6ea8bcd831c827b0321fb76bdR63-R69 added support for `coc.nvim`, if something similar is needed here a PR to update the SDK would be welcome

> Also find $$ seems to expand to what I think is the process ID

`yarn@>=3.0.0-rc.1` changed to using `__virtual__` instead of `$$virtual` (https://github.com/yarnpkg/berry/pull/2595)

tapayne88:
@merceyz are both those changes only available in v3?

merceyz:
The SDK part is generated by `@yarnpkg/pnpify@^3` and is already present in your repro https://github.com/tapayne88/yarn-pnp-demo/blob/a7188d568576ad1b16c4a43f12e00c8961e04e4c/.yarn/sdks/typescript/lib/tsserver.js#L63-L69, as for the `$$virtual` part; yes, it's a breaking change so it can't be backported

tapayne88:
Ok @merceyz, from my understanding it sounds like we'll need to implement something in the language-server to recognise these virtual file paths being emitted from tsserver and translate them into on-disk paths. That sound right or am I way off? 😄 

Sources:
https://yarnpkg.com/advanced/architecture
https://yarnpkg.com/advanced/lexicon#virtual-package

merceyz:
That is an option, but you can file a PR to Yarn that removes the virtuals when the `hostInfo` matches whatever IDE you're using (ref https://github.com/theia-ide/typescript-language-server/pull/220#issuecomment-863197308)

tapayne88:
Thanks @merceyz, yeah think I'm following. I've been hacking around and I'm a bit confused as to why we `resolve` the `zipfile` path otherwise I _think_ it might work - appreciate you might not have the context for that being a coc-nvim thing.

https://github.com/yarnpkg/berry/blob/4b349329427aebc4460d8481ae8591a8e88ff4a2/.yarn/sdks/typescript/lib/tsserver.js#L68

server/src/lsp-server.ts
```diff
@@ -123,6 +126,7 @@ export class LspServer {
 
         this.tspClient.start();
         this.tspClient.request(CommandTypes.Configure, {
+            hostInfo: 'coc-nvim',
```
tapayne88, line 129: This isn't quite right as the coc-nvim behaviour includes a `path.resolve` call on a string prepended with `zipfile:` - I can't understand how this is useful to coc-nvim but it definitely doesn't work for us

https://github.com/yarnpkg/berry/blob/4b349329427aebc4460d8481ae8591a8e88ff4a2/.yarn/sdks/typescript/lib/tsserver.js#L68

server/src/protocol-translation.ts
```diff
@@ -18,10 +18,27 @@ export function uriToPath(stringUri: string): string | undefined {
     return uri.fsPath;
 }
 
+// Identify yarn PnP archive files (.zip)
+const archivePathRE = /^\/zipfile:/;
```
tapayne88, line 22: This all feels a bit gross and not sure how it'll play with windows paths but here it is anyway 😬

server/src/modules-resolver.ts
```diff
@@ -5,20 +5,27 @@
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  */
 
-import * as fs from 'fs';
-import * as paths from 'path';
+import * as fs from "fs";
+import * as paths from "path";
```
rchl, line 9: Please don't change the quote types. Especially as the code base is predominantly using single quotes.

server/src/lsp-server.ts
rchl: It's insane how yarn-specific this fix has to be. Even to the point of disguising as a specific editor...

I'm not saying it's unacceptable but it does feel gross indeed.
And if that part is kept, I hope it doesn't break other editors or at least makes things harder for those.

tapayne88: Yeah, it does feel very yarn / vim specific passing through `zipfile:` prepended paths. We could make `hostInfo` configurable via CLI

rchl: > This isn't quite right as the coc-nvim behaviour includes a `path.resolve` call on a string prepended with `zipfile:` - I can't understand how this is useful to coc-nvim but it definitely doesn't work for us
> 
> [yarnpkg/berry@`4b34932`/.yarn/sdks/typescript/lib/tsserver.js#L68](https://github.com/yarnpkg/berry/blob/4b349329427aebc4460d8481ae8591a8e88ff4a2/.yarn/sdks/typescript/lib/tsserver.js#L68)

Through `initializationOptions` would make most sense.

tapayne88: Of course! Let me see how that looks

tapayne88:
Wonder if the handling of `.zip` files could be handled by neovim itself
https://github.com/neovim/neovim/issues/14867

server/src/protocol-translation.ts
tapayne88: I think we can drop these changes if we get a change merged into yarn that just exposes the resolved virtual module's on disk path, like this

https://github.com/tapayne88/berry/commit/ed189a9cbd97fe94e066b0a10e3b8c56d3c2b97d

server/src/protocol-translation.ts
```diff
@@ -18,8 +18,20 @@ export function uriToPath(stringUri: string): string | undefined {
     return uri.fsPath;
 }
 
+const parseURI = (filepath: string): URI => {
```
tapayne88, line 21: @rchl I think this is better now - no weird regex stuff 😄  `tsserver` will now either emit filepaths (as before) or URIs. We attempt to parse the filepath and if it's a URI we respect that and pass it upstream, otherwise turn it into a URI with a `file://` scheme

server/src/lsp-server.ts
```diff
@@ -75,8 +75,11 @@ export class LspServer {
         if (this.options.tsserverPath) {
             return this.options.tsserverPath;
         }
-        // 1) look into node_modules of workspace root
-        const executable = findPathToModule(this.rootPath(), `.bin/${getTsserverExecutable()}`);
+        // 1) look into node_modules or yarn PnP SDKs of workspace root
+        const executable = findPathToModule(this.rootPath(), [
+            path.join("node_modules", ".bin", getTsserverExecutable()),
+            path.join(".yarn", "sdks", "typescript", "lib", "tsserver.js"),
+        ])
```
rchl, line 82: We're likely gonna get https://github.com/theia-ide/typescript-language-server/pull/195 in since that one is clean and a proper way to do it. That might force you to rethink this part but I think it should be anyway because this is just too hacky. We really should be relying on the native resolver first and then add specific behavior for yarn sdk if really necessary.

tapayne88: Ok, happy to look at that. I guess we could add a new step 1 for resolving the tsserver path that checks for the yarn sdk first. I think it will need to be that way round as `require.resolve` will work for PnP but it will find the wrong file. There could be a slight performance penalty with that approach for non-yarn PnP project as they'll check for `.yarn/sdks` all the way up to `/` before using `require.resolve`.

rchl: We don't really need to exist `findPathToModule`. This logic can be separate. This only needs to be done on initialize, from one place, anyway.

That said, `findPathToModule` is only used from two places during initialize anyway so either way is fine now right now at least.

merceyz: Can't it use something like the `tsserver.tsdk` config that is generated for `coc-vim` and `vscode`?
https://github.com/yarnpkg/berry/blob/a6312e23c2cff2ee8bc338a9eaff486aa5cb97bc/.vim/coc-settings.json#L5

tapayne88: I guess we could go searching through these paths for each editor in the language server (`.vim/coc-settings.json` and `.vscode/settings.json`) or editor could do it themselves and pass through the path but then each editor / user of the language server would have to solve yarn PnP

server/src/modules-resolver.ts
```diff
@@ -12,3 +14,20 @@ export function findPathToModule(dir: string, moduleName: string): string|undefi
         return undefined;
     }
 }
+
+export function findPathToSdk(
+    dir: string,
+    moduleName: string
+): string | undefined {
```
rchl, line 21: This is specific to yarn so let's please call it what it is since SDK is a very generic term and can be confusing in context of this repo.

Also no need for such aggressive line breaking. This would do :)
```suggestion
export function findPathToYarnSdk(dir: string, moduleName: string): string | undefined {
```

server/src/modules-resolver.ts
```diff
@@ -12,3 +14,20 @@ export function findPathToModule(dir: string, moduleName: string): string|undefi
...
+    const stat = fs.statSync(dir);
+    if (stat.isDirectory()) {
+        const candidate = path.resolve(dir, '.yarn/sdks', moduleName);
```
rchl, line 24: Please separate two directories so that those are joined with proper separator:
```suggestion
        const candidate = path.resolve(dir, '.yarn', 'sdks', moduleName);
```

server/src/lsp-server.ts
```diff
@@ -75,16 +75,21 @@ export class LspServer {
...
+        if (sdk) {
+            return sdk;
+        }
+        // 2) look into node_modules of workspace root
         const executable = findPathToModule(this.rootPath(), `.bin/${getTsserverExecutable()}`);
         if (executable) {
             return executable;
         }
-        // 2) use globally installed tsserver
+        // 3) use globally installed tsserver
```
rchl, line 88: I'm thinking about a case where:
 - the current project uses npm (so no any PNP magic)
 - The user has yarn sdk initialized at `~/.yarn/sdk/...` (is that even an option?)

Assuming that that's a possible scenario, what would be an expected behavior here? I think picking yarn sdk version would be incorrect.

server/src/lsp-server.ts
rchl: Reading editor-specific configuration files is very much against the idea of LSP. I've very much opposed to that.

Also, nobody uses this server with vscode since VSCode has built-in server so I'm not sure it would be even relevant to use the option defined there.

rchl: This part was changed so resolving this issue.

server/src/protocol-translation.ts
rchl: I will think about it more. Need to check where it's used and how.

server/src/lsp-server.ts
rchl: That said, I agree that ideally the editor would do that part. Currently it's possible to pass the tsserver path through command line option (https://github.com/theia-ide/typescript-language-server/blob/49520504d5dc2b808c8ffce9457bbfa1129315b1/server/src/cli.ts#L24-L24)

server/src/modules-resolver.ts
tapayne88: 👍 sorry, my default prettier setup. Would be good to get something like prettier included in the project to avoid issues like this

server/src/lsp-server.ts
tapayne88: Yeah I think so but as I understand it yarn v2 (and PnP) is setup per project so I don't think it's possible to have a user level setup for sdks. @merceyz do you know if that is correct?

server/src/modules-resolver.ts
rchl: Personal opinion but prettier is bad since it is very opinionated without letting the user customize how things should look much. So we're not gonna introduce it. The eslint rules, even if don't force all the style, should be enough.

I think that in general you should not be running prettier on foreign projects since there is a high chance that it will just not be in line what the project uses. If I'd be using it, I'd disable it by default and only enable on a per-project case (for my projects only).

rchl:
General note: you don't have to and should really force-push to a PR branch. We're gonna likely squash everything anyway and force-pushing just makes it harder to review because I can't see the incremental changes that you are doing then.

If you really think it's beneficial to merge the changes as separate commits, we can agree on something by discussing it. Like rebasing and force-pushing only after review is approved.

server/src/modules-resolver.ts
```diff
@@ -12,3 +14,17 @@ export function findPathToModule(dir: string, moduleName: string): string|undefi
         return undefined;
     }
 }
+
+export function findPathToSdk(dir: string, moduleName: string): string | undefined {
```
rchl, line 18: Will re-create this issue to rename the function. My old one got outdated due to force-pushing.

tapayne88:
> General note: you don't have to and shouldn't really force-push to a PR branch. We're gonna likely squash everything anyway and force-pushing just makes it harder to review because I can't see the incremental changes that you are doing then.
> 
> If you really think it's beneficial to merge the changes as separate commits, we can agree on something by discussing it. Like rebasing and force-pushing only after review is approved.

Apologies for the rebasing, force of habit - completely see where you're coming from

server/src/modules-resolver.ts
tapayne88: Yeah, I usually find myself in project with prettier so it's not normally a problem - I find it a really useful tool but that's probably because I'm a bit lazy 😄 

I can totally see why people don't like the opinions of prettier but I personally think it's better to have a tool that sets the opinion rather than trying to match up with anyone's opinion. By using a tool the conversation is only "have you run prettier" and ensuring the pipeline passes.

server/src/lsp-server.ts
merceyz: Correct, there should be no need to check the home directory for the SDK

rchl: Thanks for clarifying.

Just to clarify a bit more, the current logic will check for `.yarn/sdk/...` up the directory tree, eventually ending up checking the user directory also (assuming that the project lives within the user's directory somewhere, which is likely the case).

So I'm wondering if we should short-cut that check to stop the traverse early.

Should it for example only traverse up to the first `package.json` that it finds?
Or should it in fact not traverse at all and just check within the `rootPath()`)?

tapayne88: I'm not too sure how you'd concisely express the "correct" algorithm. It _should_ traverse up until it leaves the repo, checking every site it finds a `package.json` file. Not finding a `.yarn/sdks` alongside a `package.json` isn't necessarily grounds to terminate as it could be a workspace configuration in which case there is another `package.json` higher up which could have the `.yarn/sdks` folder 😬

rchl: OK. The current behavior is probably fine then.

rchl (approved):

rchl:
Let's give it a try.

robert-zaremba:
Let's add a note, that both `typescript-language-server` and `typescript` must be installed in the project (global installation is not enough):
```
yarn add -D typescript-language-server typescript
```

Otherwise `yarn dlx @yarnpkg/sdks base` will not create the wrappers, and will create only single file:
```
tree .yarn/sdks
.yarn/sdks
└── integrations.yml
```

robert-zaremba:
Let's add a note, that both `typescript-language-server` and `typescript` must be installed in the project (global installation is not enough):
```
yarn add -D typescript-language-server typescript
```

Otherwise `yarn dlx @yarnpkg/sdks base` will not create the wrappers, and will create only single file:
```
tree .yarn/sdks
.yarn/sdks
└── integrations.yml
```

PR #221, opened 2021-06-20
Replace tslint with eslint

rchl:
The eslint rules were aligned to the current style of the project to
ensure consistency without forcing another style.

This should result in no user-facing changes


DonnieWest (approved):
LGTM

PR #226, opened 2021-06-30
Update internal Typescript dependency from 3.9.0 to 4.3.4

apexskier:
This updates the project's TypeScript version from version 3.9.0 to 4.3.4.

I'm working on this in the context of https://github.com/apexskier/nova-typescript/issues/273, because I need to update types to properly type the `skipDestructiveCodeActions` arg.

There were fewer changes than I'd expect, here are the highlights:

* Several types in the tsserver protocol are now `string | tsp.SymbolDisplayPart[]` instead of `string`. I was originally playing with improving the formatting here, but decided not to include that in the scope of this change. (although I'd like to make some improvements separately)
* There are several new command types, so I've updated `server/src/tsp-command-types.ts`

Along with automated tests, I've been manually testing this underneath https://github.com/apexskier/nova-typescript and it seems to be working the same as before.

### DIFF
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -163,10 +163,10 @@ function asCodeActions(details: tsp.CompletionEntryDetails, filepath: string): {
     // Also check if we still have to apply other workspace edits and commands
     // using a vscode command
     const additionalTextEdits: lsp.TextEdit[] = [];
-    let hasReaminingCommandsOrEdits = false;
+    let hasRemainingCommandsOrEdits = false;
     for (const tsAction of details.codeActions) {
         if (tsAction.commands) {
-            hasReaminingCommandsOrEdits = true;
+            hasRemainingCommandsOrEdits = true;
         }
 
         // Apply all edits in the current file using `additionalTextEdits`
@@ -177,14 +177,14 @@ function asCodeActions(details: tsp.CompletionEntryDetails, filepath: string): {
                         additionalTextEdits.push(toTextEdit(textChange));
                     }
                 } else {
-                    hasReaminingCommandsOrEdits = true;
+                    hasRemainingCommandsOrEdits = true;
                 }
             }
         }
     }
 
     let command: lsp.Command | undefined = undefined;
-    if (hasReaminingCommandsOrEdits) {
+    if (hasRemainingCommandsOrEdits) {
         // Create command that applies all edits not in the current file.
         command = {
             title: '',
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -6,7 +6,7 @@
  */
 
 import * as lsp from 'vscode-languageserver';
-import tsp from 'typescript/lib/protocol';
+import type tsp from 'typescript/lib/protocol';
 import { LspDocument } from './document';
 import { ScriptElementKind } from './tsp-command-types';
 import { asRange, toTextEdit, asPlainText, asDocumentation } from './protocol-translation';
@@ -16,7 +16,7 @@ export interface TSCompletionItem extends lsp.CompletionItem {
     data: tsp.CompletionDetailsRequestArgs;
 }
 
-export function asCompletionItem(entry: import('typescript/lib/protocol').CompletionEntry, file: string, position: lsp.Position, document: LspDocument): TSCompletionItem {
+export function asCompletionItem(entry: tsp.CompletionEntry, file: string, position: lsp.Position, document: LspDocument): TSCompletionItem {
     const item: TSCompletionItem = {
         label: entry.name,
         kind: asCompletionItemKind(entry.kind),
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -70,7 +70,7 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
     return item;
 }
 
-export function asCompletionItemKind(kind: ScriptElementKind): lsp.CompletionItemKind {
+function asCompletionItemKind(kind: ScriptElementKind): lsp.CompletionItemKind {
     switch (kind) {
         case ScriptElementKind.primitiveType:
         case ScriptElementKind.keyword:
@@ -114,7 +114,7 @@ export function asCompletionItemKind(kind: ScriptElementKind): lsp.CompletionIte
     return lsp.CompletionItemKind.Property;
 }
 
-export function asCommitCharacters(kind: ScriptElementKind): string[] | undefined {
+function asCommitCharacters(kind: ScriptElementKind): string[] | undefined {
     const commitCharacters: string[] = [];
     switch (kind) {
         case ScriptElementKind.memberGetAccessorElement:
@@ -152,7 +152,7 @@ export function asResolvedCompletionItem(item: TSCompletionItem, details: tsp.Co
     return item;
 }
 
-export function asCodeActions(details: tsp.CompletionEntryDetails, filepath: string): {
+function asCodeActions(details: tsp.CompletionEntryDetails, filepath: string): {
     command?: lsp.Command; additionalTextEdits?: lsp.TextEdit[];
 } {
     if (!details.codeActions || !details.codeActions.length) {
@@ -203,7 +203,7 @@ export function asCodeActions(details: tsp.CompletionEntryDetails, filepath: str
     };
 }
 
-export function asDetail({ displayParts, source }: tsp.CompletionEntryDetails): string | undefined {
+function asDetail({ displayParts, source }: tsp.CompletionEntryDetails): string | undefined {
     const result: string[] = [];
     const importPath = asPlainText(source);
     if (importPath) {
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -203,11 +203,11 @@ function asCodeActions(details: tsp.CompletionEntryDetails, filepath: string): {
     };
 }
 
-function asDetail({ displayParts, source }: tsp.CompletionEntryDetails): string | undefined {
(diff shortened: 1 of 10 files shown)


server/src/utils.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 export class Deferred<T> {
-    resolve: (value?: T) => void;
+    resolve: (value?: T | PromiseLike<T>) => void;
```
rchl, line 9: In the updated Promise types the value is no longer optional so this should just be:
```suggestion
    resolve: (value: T) => void;
```

server/src/tsp-client.ts
```diff
@@ -10,10 +10,9 @@ import * as path from 'path';
 import * as cp from 'child_process';
 import * as readline from 'readline';
 import * as decoder from 'string_decoder';
-import protocol from 'typescript/lib/protocol';
+import tsp from 'typescript/lib/protocol';
```
rchl, line 13: I'd suggest this instead:
```suggestion
import protocol, { CommandTypes } from 'typescript/lib/protocol';
```
Then you avoid all unnecessary changes and the diff is minimal.

Similarly in other files.

rchl (changes requested):

server/src/protocol-translation.ts
```diff
@@ -233,7 +238,15 @@ export function asTagsDocumentation(tags: tsp.JSDocTagInfo[]): string {
...
         case 'param': {
-            const body = (tag.text || '').split(/^([\w.]+)\s*-?\s*/);
+            // TODO: What does this actually look like? Is the original logic
+            // just doing something that the new type should do?
+            let formattedTag = '';
+            if (typeof tag.text === 'string') {
+                formattedTag = tag.text;
+            } else if (tag.text) {
+                formattedTag = tag.text.map(symbol => `${symbol.text} (${symbol.kind})`).join(', ');
+            }
```
rchl, line 248: I'd like something like:
```suggestion
const formattedTag = typeof tag.text === 'string' ? tag.text : asPlainText(tag.text);
```

and then update `asPlainText` to handle the `kind` in `SymbolDisplayPart`.

rchl (changes requested):

server/src/protocol-translation.ts
rchl: But to be fair, looking at the description of `kind` which is:

> The symbol's kind (such as 'className' or 'parameterName' or plain 'text')

I don't think we would necessarily want to to show the kind value. Certainly not when it's `text` but even in other cases it's probably not needed (at least for now).

So it seems fine to just do:

```
const formattedTag = typeof tag.text === 'string' ? tag.text : asPlainText(tag.text);
```
for now without updating `asPlainText`.

apexskier: Okay, I'm probably fine with that. I'm still playing with this, I want to see what it ends up looking like in https://github.com/apexskier/nova-typescript/ with full information.

rchl: Cool, please let me not (screenshots would be nice :)).

BTW. To enable typescript to send `SymbolDisplayPart` I guess some flag/setting needs to be enabled?

rchl: Though note the documentation that I've referenced above:

> The symbol's kind (such as 'className' or 'parameterName' or plain 'text')

Those kinds look too low-level to show in the UI anyway. If we would want to show those we probably would have to map them to UI-friendly strings.

apexskier: Looks like it's controlled by `UserPreferences.displayPartsForJSDoc`
https://github.com/microsoft/TypeScript/blob/62f9155a9d3c734854f3340e2b74fb799bf407fe/src/server/protocol.ts#L989

apexskier: At time of writing with the TODO experimental code, here's an example of what this looks like (obviously lots of work to be done)

TypeScript Extension – index d ts - Nova - 2021-07-02 at 21 34 45
TypeScript Extension – novaUtils ts - Nova - 2021-07-02 at 21 34 58

rchl: Right. So that `text` kind looks very redundant so far. I wonder in which cases we would get the other "kind"s.

server/src/protocol-translation.ts
```diff
@@ -243,7 +255,7 @@ export function asTagDocumentation(tag: tsp.JSDocTagInfo): string {
                 }
                 return label + (doc.match(/\r\n|\n/g) ? '  \n' + doc : ` — ${doc}`);
             }
-            // fall-through
```
apexskier, line 246: There's no subsequent logic in the `switch` to fall through to.

server/src/protocol-translation.ts
apexskier: FYI, I'm removing this in this initial PR. I'll do a secondary one to change the formatting behavior.

server/src/document-symbol.ts
```diff
@@ -55,7 +55,7 @@ function collectDocumentSymbolsInRange(parent: tsp.NavigationTree, symbols: lsp.
     return shouldInclude;
 }
 
-export function collectSymbolInformations(uri: string, current: tsp.NavigationTree, symbols: lsp.SymbolInformation[], containerName?: string): boolean {
```
apexskier, line 58: Typo fix

server/src/lsp-server.spec.ts
```diff
@@ -491,106 +491,58 @@ describe('code actions', () => {
 
         assert.deepEqual(result, [
```
apexskier, line 492: TS 4 has different quick-actions. Note that this won't change if someone's running the language server on an older typescript version, since these are pulled from the tslib.

rchl:
I've updated the lock file on master after previous changes (seems like that was forgotten before) so there is a conflict.

rchl:
I hope you don't mind that I'm gonna push some minor changes to your branch.

server/src/lsp-server.ts
```diff
@@ -487,7 +485,13 @@ export class LspServer {
             { language: 'typescript', value: result.body.displayString }
         ];
         const tags = asTagsDocumentation(result.body.tags);
-        contents.push(result.body.documentation + (tags ? '\n\n' + tags : ''));
+        let documentation: string;
+        if (typeof result.body.documentation === 'string') {
+            documentation = result.body.documentation;
+        } else {
+            documentation = result.body.documentation?.map(p => p.text).join('') || '';
```
rchl, line 492: According to the types this can't be undefined so:
```suggestion
            documentation = result.body.documentation.map(p => p.text).join('') || '';
```

But can you instead use the `asPlainText` helper function everywhere instead?

server/src/protocol-translation.ts
```diff
@@ -233,7 +211,13 @@ export function asTagsDocumentation(tags: tsp.JSDocTagInfo[]): string {
 export function asTagDocumentation(tag: tsp.JSDocTagInfo): string {
     switch (tag.name) {
         case 'param': {
-            const body = (tag.text || '').split(/^([\w.]+)\s*-?\s*/);
+            let text: string;
+            if (typeof tag.text === 'string') {
+                text = tag.text;
+            } else {
+                text = tag.text?.map(p => p.text).join('') || '';
```
rchl, line 218: Can use `asPlainText`?

server/src/protocol-translation.ts
```diff
@@ -261,17 +245,24 @@ export function asTagBodyText(tag: tsp.JSDocTagInfo): string | undefined {
         return undefined;
     }
 
+    let text: string;
+    if (typeof tag.text === 'string') {
+        text = tag.text;
+    } else {
+        text = tag.text.map(p => p.text).join('');
```
rchl, line 252: Can use `asPlainText`?

rchl (changes requested):

server/src/protocol-translation.ts
```diff
@@ -233,7 +210,16 @@ export function asTagsDocumentation(tags: tsp.JSDocTagInfo[]): string {
...
-            const body = (tag.text || '').split(/^([\w.]+)\s*-?\s*/);
+            if (!tag.text) {
+                break;
+            }
+            let text: string;
+            if (typeof tag.text === 'string') {
+                text = tag.text;
+            } else {
+                text = asPlainText(tag.text);
+            }
```
rchl, line 221:
```suggestion
            const text = asPlainText(tag.text);
```

server/src/completion.ts
```diff
@@ -203,11 +203,11 @@ export function asCodeActions(details: tsp.CompletionEntryDetails, filepath: str
...
 
-export function asDetail({ displayParts, source }: tsp.CompletionEntryDetails): string | undefined {
+function asDetail({ displayParts, sourceDisplay, source: deprecatedSource }: tsp.CompletionEntryDetails): string | undefined {
     const result: string[] = [];
-    const importPath = asPlainText(source);
-    if (importPath) {
-        result.push(`Auto import from '${importPath}'`);
+    const source = sourceDisplay || deprecatedSource;
+    if (source) {
+        result.push(`Auto import from '${asPlainText(source)}'`);
```
rchl, line 210: Any cases where this change has observable effect?

apexskier: I haven't found any, no. It should be backwards compatible with old versions of typescript since I'm still using `source`, but I didn't test explicitly.

rchl: Well, it looks like a good change anyway as `sourceDisplay` is supposed to be the human-readable one anyway.

rchl (approved):
Great stuff. Thanks.

PR #227, opened 2021-07-03
Mark completion items as deprecated if JSDoc says so

apexskier:
Just a simple new feature.


server/src/completion.ts
```diff
@@ -148,10 +148,15 @@ export function asCommitCharacters(kind: ScriptElementKind): string[] | undefine
 export function asResolvedCompletionItem(item: TSCompletionItem, details: tsp.CompletionEntryDetails): TSCompletionItem {
     item.detail = asDetail(details);
     item.documentation = asDocumentation(details);
+    item.deprecated = isDeprecated(details.tags);
```
rchl, line 151: Ironic because `CompletionItem.deprecated` is deprecated itself. The spec says to use `tags` instead:

> @deprecated Use `tags` instead if supported.

rchl: I understand that supporting `tags` would probably require more work (checking client capability, maybe also updating vscode libraries?) so I'm willing to accept this for now if you think it's too involved for now.

server/src/completion.ts
```diff
@@ -148,10 +148,15 @@ export function asCommitCharacters(kind: ScriptElementKind): string[] | undefine
...
     Object.assign(item, asCodeActions(details, item.data.file));
     return item;
 }
 
+function isDeprecated(tags?: tsp.JSDocTagInfo[]): boolean | undefined {
+    return tags?.some(tag => tag.name.toLowerCase() === 'deprecated');
+}
```
rchl, line 158: I'm not sure what's the prevalent convention in this code base as I haven't had time to explore it much yet but I would probably prefer tighter typing (`tags` parameter non-optional, return type only `boolean`).

Not a blocker but consider that instead please.

server/src/completion.ts
apexskier: Whoops, I didn't notice that since I was just looking at the current dependency chain. I'll look into tags.

server/src/completion.ts
apexskier: I'm using this in a context where the tags are optional on the object, and I like the syntax of making the argument optional.

Instead of `obj.tags ? isDeprecated(obj.tags) : undefined`, I can just do `isDeprecated(obj.tags)`

server/src/lsp-server.spec.ts
```diff
@@ -99,6 +100,40 @@ describe('completion', () => {
...
+            textDocument: doc
+        });
+        const pos = position(doc, 'foo(); // call me');
+        const proposals = await server.completion({
+            textDocument: doc,
+            position: pos
+        }) as TSCompletionItem[];
+        const item = proposals.filter(i => i.label === 'foo')[0];
+        const resolvedItem = await server.completionResolve(item);
+        assert.isTrue(resolvedItem.detail !== undefined, JSON.stringify(resolvedItem, undefined, 2));
```
rchl, line 130: Why would we stringify `undefined` if that assertion fails? Some debug left over code?

Also `assert.isDefined` is a better option here.

server/src/lsp-server.spec.ts
```diff
@@ -99,6 +100,40 @@ describe('completion', () => {
...
+        };
+        server.didOpenTextDocument({
+            textDocument: doc
+        });
+        const pos = position(doc, 'foo(); // call me');
+        const proposals = await server.completion({
+            textDocument: doc,
+            position: pos
+        }) as TSCompletionItem[];
+        const item = proposals.filter(i => i.label === 'foo')[0];
```
rchl, line 128: `find` would look cleaner

server/src/lsp-server.spec.ts
apexskier: Ah, I'm not used to chai and was looking for `assert.isDefined`. This whole line is copied from another test, so it's old.

server/src/completion.ts
rchl: Yeah, maybe at least lets make the return type be just a boolean.

server/src/lsp-server.spec.ts
```diff
@@ -50,10 +50,15 @@ describe('completion', () => {
...
             position: pos
         }) as TSCompletionItem[];
-        assert.isTrue(proposals.length > 800, String(proposals.length));
-        const item = proposals.filter(i => i.label === 'addEventListener')[0];
+        assert.isAtLeast(proposals.length, 800);
+        const item = proposals.find(i => i.label === 'addEventListener');
+        if (!item) {
+            assert.fail('missing addEventListener completion');
+            return;
+        }
```
rchl, line 58: Or:
```suggestion
        assert.isDefined(item, 'missing addEventListener completion');
        if (!item) {
            return;
        }
```
(annoying that TS doesn't understand this automatically ;))

server/src/completion.ts
rchl: It looks like it would actually require an update of vscode libraries though so probably should do that in a separate PR first.

server/src/completion.ts
rchl: It looks like it would actually require an update of vscode libraries though so probably should do that in a separate PR first.

server/src/completion.ts
```diff
@@ -148,10 +148,15 @@ export function asCommitCharacters(kind: ScriptElementKind): string[] | undefine
 export function asResolvedCompletionItem(item: TSCompletionItem, details: tsp.CompletionEntryDetails): TSCompletionItem {
     item.detail = asDetail(details);
     item.documentation = asDocumentation(details);
+    item.deprecated = isDeprecated(details.tags);
     Object.assign(item, asCodeActions(details, item.data.file));
     return item;
 }
 
+function isDeprecated(tags?: tsp.JSDocTagInfo[]): boolean | undefined {
+    return tags?.some(tag => tag.name.toLowerCase() === 'deprecated');
```
predragnikolic, line 157: If I recall correctly, you can check if a TypeScript completion item is deprecated by checking the `TSCompletionItem.kindModifiers` (since version 4).

https://github.com/microsoft/TypeScript/pull/38523/files#diff-ac54f9219619989a94c776b937ce1ebcbcc7dc6d7c424effab113f2ed0eb016fR19

You can see this [PR](https://github.com/microsoft/TypeScript/pull/38523/files#diff-ac54f9219619989a94c776b937ce1ebcbcc7dc6d7c424effab113f2ed0eb016fR19) that added the deprecated support in TS for more info. 

-----
I think that it would be good to rely on `kindModifiers` for workspace that has TS version 4 and above.

For workspace(project) that use a version of typescript below version 4. 
I would not try to put additional effort to mark the completion item as deprecated (because TS didn't support it prior version 4).

But If you really want to mark completion items as deprecated, you can use this approach that you already implemented. (by checking the tags in the docs comment).

predragnikolic: I haven't tried this PR, 

but I think performance would be affected when trying to traverse the docs comment for each completion item(CI). So the completion response might need (a lot more, depending on the project) time to compute and return. (performance was also a [concern](https://github.com/microsoft/TypeScript/pull/38523#issuecomment-645528105) when they implemented it in TS)

That is the reason why I would rely only on `kindModifiers`.

rchl: Changed the code to use `kindModifiers` and without having to resolve item.

server/src/completion.ts
rchl: You've updated the vscode libaries and I've changed to not set the `deprecated` property so resolving.

rchl:
@apexskier I've updated the code to address the review comments. Setting the `deprecated` tag based on `kindModifiers`. The `kindModifiers` is available without having to resolve the completion so that's a bonus also.

rchl:
@apexskier if you are not too busy these days, would you be able to review my changes?

PR #228, opened 2021-07-03
feat: allow skip destructive actions on running OrganizeImports

apexskier:
This adds support for the new `skipDestructiveCodeActions` argument to TypeScript's organize imports feature - https://github.com/microsoft/TypeScript/issues/43051 to support https://github.com/apexskier/nova-typescript/issues/273.

~This PR is built on top of #226, and will be rebased when that's merged. For now you can see the diff at https://github.com/apexskier/typescript-language-server/compare/typescript-4...apexskier:organize-imports-skip-destructive-actions.~ To properly type and types this, it's blocked on upgrading to typescript 4.4, which is still in beta.

Support is added in two places:

1. Automatically inferring the proper value based on diagnostics for the file when returning code actions.
2. Supporting sending it when manually executing the organize imports action.

I've also added documentation to the readme about the supported commands that can be manually executed.


server/src/lsp-server.ts
```diff
@@ -698,8 +698,17 @@ export class LspServer {
...
+            const skipDestructiveCodeActions = params.context.diagnostics.some(
+                // assume no severity is an error
+                d => (d.severity ?? 0) <= 2
+            );
             actions.push(...provideOrganizeImports(
-                await this.getOrganizeImports({ scope: { type: 'file', args } })
+                await this.getOrganizeImports({
+                    scope: { type: 'file', args },
+                    skipDestructiveCodeActions
+                })
```
rchl, line 711: Do we need to start version-checking here or older versions will accept this argument with no complains?

server/src/lsp-server.ts
```diff
@@ -767,11 +776,15 @@ export class LspServer {
...
+            const additionalArguments = arg.arguments[1] as undefined | {
+                skipDestructiveCodeActions?: boolean;
+            };
             const { body } = await this.tspClient.request(CommandTypes.OrganizeImports, {
                 scope: {
                     type: 'file',
                     args: { file }
-                }
+                },
+                ...additionalArguments
```
rchl, line 787: I would like to be more "defensive" here and only pass the properties that we expect here. Otherwise this can go bad at runtime.

rchl:
@apexskier are you planning to pick it up again or you'd prefer ne to take over?

rchl (approved):

Issue #191, opened 2021-03-11
doRequestDiagnostics of LspServer looks did not handle err? no 'catch'?


Shane-XB-Qian:
err got from lsp client:

```log
node:events:355
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:160:15)
    at writeGeneric (node:internal/stream_base_commons:151:3)
    at Socket._writeGeneric (node:net:773:11)
    at Socket._write (node:net:785:8)
    at writeOrBuffer (node:internal/streams/writable:395:12)
    at Socket.Writable.write (node:internal/streams/writable:340:10)
    at TspClient.sendMessage (/usr/local/lib/node_modules/typescript-language-server/lib/tsp-client.js:100:33)
    at TspClient.request (/usr/local/lib/node_modules/typescript-language-server/lib/tsp-client.js:68:14)
    at LspServer.<anonymous> (/usr/local/lib/node_modules/typescript-language-server/lib/lsp-server.js:195:45)
    at Generator.next (<anonymous>)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:188:8)
    at emitErrorCloseNT (node:internal/streams/destroy:153:3)
    at processTicksAndRejections (node:internal/process/task_queues:81:21) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}
```

looks doRequestDiagnostics of LspServer looks did not handle err? no 'catch'?
```js
    doRequestDiagnostics() {
        return __awaiter(this, void 0, void 0, function* () {
            this.cancelDiagnostics();
            const geterrTokenSource = new lsp.CancellationTokenSource();
            this.diagnosticsTokenSource = geterrTokenSource;
            const { files } = this.documents;
            try {
                return yield this.tspClient.request("geterr" /* Geterr */, { delay: 0, files }, this.diagnosticsTokenSource.token);
            }
            finally {
                if (this.diagnosticsTokenSource === geterrTokenSource) {
                    this.diagnosticsTokenSource = undefined;
                }
            }
        });
    }
```

// installed from npm.

latk:
A similar error occurred in #163, but without resolution.

Shane-XB-Qian:
```
dist-tags:
dev: 0.4.0-dev.42dcf0d    latest: 0.5.4             next: 0.4.0-next.95a0000  

published 22 minutes ago
```

[...]

PR #229, opened 2021-07-03
Only publish diagnostics if client supports the capability

rchl:
Ensure that diagnostics are only published if client supports the
textDocument.publishDiagnostics capability.

Relates to #191

### DIFF
--- a/server/src/lsp-server.spec.ts
+++ b/server/src/lsp-server.spec.ts
@@ -30,6 +30,10 @@ beforeEach(() => {
     server.closeAll();
 });
 
+after(() => {
+    server.closeAll();
+});
+
 describe('completion', () => {
     it('simple test', async () => {
         const doc = {
@@ -804,12 +808,12 @@ export function factory() {
 describe('diagnostics (no client support)', () => {
     before(async () => {
         // Remove the "textDocument.publishDiagnostics" client capability.
-        const clientCapabilitesOverride = getDefaultClientCapabilities();
-        delete clientCapabilitesOverride.textDocument?.publishDiagnostics;
+        const clientCapabilitiesOverride = getDefaultClientCapabilities();
+        delete clientCapabilitiesOverride.textDocument?.publishDiagnostics;
         server = await createServer({
             rootUri: null,
             publishDiagnostics: args => diagnostics.push(args),
-            clientCapabilitesOverride
+            clientCapabilitiesOverride
         });
     });
 
@@ -832,6 +836,6 @@ describe('diagnostics (no client support)', () => {
         await server.requestDiagnostics();
         await new Promise(resolve => setTimeout(resolve, 200));
         const diagnosticsForThisFile = diagnostics.filter(d => d!.uri === doc.uri);
-        assert.equal(diagnosticsForThisFile.length, 0, 'Unexpected diagnostics received');
+        assert.isEmpty(diagnosticsForThisFile, 'Unexpected diagnostics received');
     }).timeout(10000);
 });
--- a/server/src/lsp-server.spec.ts
+++ b/server/src/lsp-server.spec.ts
@@ -9,7 +9,7 @@ import * as chai from 'chai';
 import * as lsp from 'vscode-languageserver';
 import * as lspcalls from './lsp-protocol.calls.proposed';
 import { LspServer } from './lsp-server';
-import { uri, createServer, position, lastPosition, filePath } from './test-utils';
+import { uri, createServer, position, lastPosition, filePath, getDefaultClientCapabilities } from './test-utils';
 import { TextDocument } from 'vscode-languageserver';
 import { TSCompletionItem } from './completion';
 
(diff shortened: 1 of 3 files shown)


rchl:
@apexskier since you are so active with the project, maybe you want to have a look at this one and get it of @DonnieWest's chest. :)

server/src/lsp-server.spec.ts
```diff
@@ -845,3 +845,38 @@ export function factory() {
...
+        };
+        server.didOpenTextDocument({
+            textDocument: doc
+        });
+
+        server.requestDiagnostics();
+        await server.requestDiagnostics();
+        await new Promise(resolve => setTimeout(resolve, 200));
+        const diagnosticsForThisFile = diagnostics.filter(d => d!.uri === doc.uri);
+        assert.equal(diagnosticsForThisFile.length, 0, 'Unexpected diagnostics received');
```
apexskier, line 880:
```suggestion
        assert.isEmpty(diagnosticsForThisFile, 'Unexpected diagnostics received');
```

server/src/lsp-server.spec.ts
```diff
@@ -845,3 +845,38 @@ export function factory() {
         );
     }).timeout(10000);
 });
+
+describe('diagnostics (no client support)', () => {
+    before(async () => {
+        // Remove the "textDocument.publishDiagnostics" client capability.
+        const clientCapabilitesOverride = getDefaultClientCapabilities();
+        delete clientCapabilitesOverride.textDocument?.publishDiagnostics;
+        server = await createServer({
```
apexskier, line 854: I'm not super familiar with all this, but is there a way to close these servers? As is, I think all the ones created will still be active until the entire test suite's process ends, and will contend for the same log files.

Not a blocking comment.

server/src/lsp-server.spec.ts
```diff
@@ -845,3 +845,38 @@ export function factory() {
...
+        }
+      `
+        };
+        server.didOpenTextDocument({
+            textDocument: doc
+        });
+
+        server.requestDiagnostics();
+        await server.requestDiagnostics();
+        await new Promise(resolve => setTimeout(resolve, 200));
```
apexskier, line 878: I don't love having a fixed timeout constant here, but I did some testing locally and it looks like it's not easy to avoid, and it's an existing pattern. (might relate to a debounce internally, or the inherent slowness in an external process)

server/src/lsp-server.ts
```diff
@@ -944,7 +947,9 @@ export class LspServer {
         if (event.event === EventTypes.SementicDiag ||
             event.event === EventTypes.SyntaxDiag ||
             event.event === EventTypes.SuggestionDiag) {
-            this.diagnosticQueue.updateDiagnostics(event.event, event as tsp.DiagnosticEvent);
+            if (this.diagnosticQueue) {
+                this.diagnosticQueue.updateDiagnostics(event.event, event as tsp.DiagnosticEvent);
+            }
```
apexskier, line 952:
```suggestion
            this.diagnosticQueue?.updateDiagnostics(event.event, event as tsp.DiagnosticEvent);
```

server/src/test-utils.ts
```diff
@@ -46,6 +57,7 @@ export async function createServer(options: {
     rootUri: string | null;
     tsserverLogVerbosity?: string;
     publishDiagnostics: (args: lsp.PublishDiagnosticsParams) => void;
+    clientCapabilitesOverride?: lsp.ClientCapabilities;
```
apexskier, line 60:
```suggestion
    clientCapabilitiesOverride?: lsp.ClientCapabilities;
```

server/src/lsp-server.spec.ts
```diff
@@ -845,3 +845,38 @@ export function factory() {
         );
     }).timeout(10000);
 });
+
+describe('diagnostics (no client support)', () => {
+    before(async () => {
+        // Remove the "textDocument.publishDiagnostics" client capability.
+        const clientCapabilitesOverride = getDefaultClientCapabilities();
```
apexskier, line 852:
```suggestion
        const clientCapabilitiesOverride = getDefaultClientCapabilities();
```

server/src/lsp-server.spec.ts
rchl: The `beforeEach` hook closes the server: https://github.com/theia-ide/typescript-language-server/blob/5b3c3d9110140ab11617d3c5586498d8a18f7215/server/src/lsp-server.spec.ts#L28-L31

Those nothing seems to be closing the server after all tests finish.

server/src/lsp-server.spec.ts
rchl: Yes, just something that was used already and likely due to debounce.
A more explicit way would be exposing the debounce timeout and use it here.
Or just disabled debouncing for tests but then the logic would be different in tests than in production.

apexskier (approved):

rchl:
As mentioned in #191, there is a client out there that doesn't announce `textDocument.publishDiagnostics` capability but actually supports diagnostics. So this would be a breaking change in some cases...
I could say that the client is just broken and should be fixed but on the other hand there is my pragmatic side...

apexskier:
Personally, I think that's acceptable. Continually supporting non-standard use cases means more maintenance cost and complexity in this codebase and disincentivizes clients from conforming properly. Given this is still v0 and is open-source, I don't think there's necessarily an obligation to fix this, but it's not an easy call to make.

rchl:
Good point

strager:
I am writing an LSP client. I noticed a change in typescript-language-server's behavior between versions 0.5.4 and 0.6.0. I think this PR is to blame.

> Continually supporting non-standard use cases means more maintenance cost and complexity in this codebase and disincentivizes clients from conforming properly. 

My understanding is that clients are **required** to support `textDocument/publishDiagnostics` [1]. From [the current spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textdocumentclientcapabilities):

> Client capabilities got introduced with version 3.0 of the protocol. They therefore only describe capabilities that got introduced in 3.x or later. Capabilities that existed in the 2.x version of the protocol are still mandatory for clients. Clients cannot opt out of providing them. So even if a client omits the ClientCapabilities.textDocument.synchronization it is still required that the client provides text document synchronization (e.g. open, changed and close notifications).

`textDocument/publishDiagnostics` has existed [since protocol version 1.x](
https://github.com/microsoft/language-server-protocol/blob/9491829c0479c736147eae1642b374846a5f1028/versions/protocol-1-x.md#publishdiagnostics-notification) and still existed in [protocol version 2.x](https://github.com/microsoft/language-server-protocol/blob/main/versions/protocol-2-x.md#publishdiagnostics-notification). My understanding is that `textDocument/publishDiagnostics` is grandfathered in, therefore client capabilities are not required to advertise client support.

[1] By "support", I mean at least tolerate `textDocument/publishDiagnostics` notifications, even if they are ignored.

rchl:
Good point but do you expect this server to change that? Is it not an option to add the capability in your client?

strager:
> Good point but do you expect this server to change that?

Yes.

> Is it not an option to add the capability in your client?

I did add a workaround to my client.

rchl:
Created a fix in https://github.com/typescript-language-server/typescript-language-server/pull/261

PR #230, opened 2021-07-04
Add support for "unnecessary" and "deprecated" diagnostic tags

rchl:
Needed to pass client capability down the chain so that we don't add tags when client doesn't support them.

There are more fields on diagnostics that are only supported in later versions (not supported in current code base) so this will also be useful in the future.

@apexskier

### DIFF
--- a/server/src/lsp-server.spec.ts
+++ b/server/src/lsp-server.spec.ts
@@ -174,7 +174,6 @@ describe('diagnostics', () => {
 
         await server.requestDiagnostics();
         await new Promise(resolve => setTimeout(resolve, 200));
-        // assert.deepEqual(JSON.stringify(diagnostics, null, 2), '');
         const resultsForFile = diagnostics.find(d => d!.uri === doc.uri);
         assert.isDefined(resultsForFile);
         const fileDiagnostics = resultsForFile!.diagnostics;
--- a/server/src/lsp-server.spec.ts
+++ b/server/src/lsp-server.spec.ts
@@ -15,7 +15,7 @@ import { TSCompletionItem } from './completion';
 
 const assert = chai.assert;
 
-let diagnostics: Array<lsp.PublishDiagnosticsParams | undefined>;
+let diagnostics: Array<lsp.PublishDiagnosticsParams>;
 
 let server: LspServer;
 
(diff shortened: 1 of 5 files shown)


server/src/lsp-server.spec.ts
```diff
@@ -155,6 +155,41 @@ describe('diagnostics', () => {
...
+        foo();
+      `
+        };
+        server.didOpenTextDocument({
+            textDocument: doc
+        });
+
+        await server.requestDiagnostics();
+        await new Promise(resolve => setTimeout(resolve, 200));
+        // assert.deepEqual(JSON.stringify(diagnostics, null, 2), '');
```
apexskier: Remove this?

server/src/lsp-server.spec.ts
```diff
@@ -155,6 +155,41 @@ describe('diagnostics', () => {
...
+        const resultsForFile = diagnostics.find(d => d!.uri === doc.uri);
```
apexskier, line 177:
```suggestion
        const resultsForFile = diagnostics.find(d => d?.uri === doc.uri);
```

server/src/lsp-server.spec.ts
```diff
@@ -155,6 +155,41 @@ describe('diagnostics', () => {
...
+        assert.isDefined(resultsForFile);
+        const fileDiagnostics = resultsForFile!.diagnostics;
+        assert.equal(fileDiagnostics.length, 2);
+        const unusedDiagnostic = fileDiagnostics.find(d => d.code === 6133);
+        assert.isDefined(unusedDiagnostic);
+        assert.deepEqual(unusedDiagnostic!.tags, [lsp.DiagnosticTag.Unnecessary]);
```
apexskier, line 183: If you wanted to avoid the null assertions, this should be safe (but type checking of the params might fail?) - this will definitely fail when run if `undefined`.

```suggestion
        assert.deepEqual(unusedDiagnostic?.tags, [lsp.DiagnosticTag.Unnecessary]);
```

apexskier (commented):
Nice!

server/src/lsp-server.spec.ts
rchl: I think it should be fine to remove `undefined` type form the `diagnostics` array instead.

server/src/lsp-server.spec.ts
rchl: This should be safe because the `assert.isDefined(unusedDiagnostic);` will fail the test before that line when `unusedDiagnostic` is undefined.

apexskier: Yeah, I just don't like leaving `!`s

apexskier (approved):

server/src/lsp-server.spec.ts
rchl: I think in this case it's very appropriate because we are expressing the fact that we know that it's defined at this point.

PR #231, opened 2021-07-04
Upgrade vscode-languageserver

apexskier:
This upgrades the core `vscode-langaugeserver` package and the LSP in use to `v3.16.0`. Relates to https://github.com/theia-ide/typescript-language-server/pull/227#discussion_r663515395.

I also pulled in vscode-languageserver-textdocument. The language server's TextDocument has been deprecated and replaced with this package.

I spent a lot of time getting the example dir working. Because the monaco language server package is not maintained much any more, it's incompatible with the newer version of the language server protocol (even when upgraded) and that leads to needing to pin different versions. I attempted a few different upgrades and hit various typescript and webpack errors along the way. I ended up deciding to just get the example working (and added some better instructions about how to verify this) and skip upgrading anything inside it.

### DIFF
--- a/example/src/client.ts
+++ b/example/src/client.ts
@@ -4,6 +4,7 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  */
+import * as monaco from 'monaco-editor'
 import { listen, MessageConnection } from 'vscode-ws-jsonrpc';
 import {
     MonacoLanguageClient, CloseAction, ErrorAction,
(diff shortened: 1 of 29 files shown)


example/src/client.ts
```diff
@@ -57,9 +54,14 @@ listen({
         // create and start the language client
         const languageClient = createLanguageClient(connection);
         languageClient.onReady().then(() => {
-            languageClient.onRequest(TypeScriptRenameRequest.type, params => {
```
apexskier, line 60: I didn't want to remove this, but because of the different language server protocol types it's now failing type checking.

predragnikolic:
Has anyone considered deleting the `/example` directory?
That would ease the maintenance of this project.

rchl (approved):

rchl:
> Has anyone considered deleting the `/example` directory?
> That would ease the maintenance of this project.

+1 from me on getting it removed.

I don't know if anyone finds that example useful. In theory it's good to have some example editor implementation but I never found a use case for it myself.

Thanks for putting effort into that part too and we'll keep it on live support for now with the possibility of removal in the future.

PR #233, opened 2021-07-05
Don't pass deprecated options to Completion request

rchl:
The deprecated options "includeExternalModuleExports" and "includeInsertTextCompletions" are now passed as global, non-deprecated typescript preferences.

To achieve that, added those as default preferences to the preferences object that the user can override.

Also switch to non-deprecated "CompletionInfo" request and handle "isIncomplete" flag (it is used when suggesting global import completions when `includeCompletionsForImportStatements` and `includeCompletionsWithInsertText` typescript preferences are enabled) and fix some Completion-related types (use LSP types instead of TSC types).

@apexskier

### DIFF
--- a/server/src/completion.ts
+++ b/server/src/completion.ts
@@ -145,7 +145,7 @@ function asCommitCharacters(kind: ScriptElementKind): string[] | undefined {
     return commitCharacters.length === 0 ? undefined : commitCharacters;
 }
 
-export function asResolvedCompletionItem(item: TSCompletionItem, details: tsp.CompletionEntryDetails): TSCompletionItem {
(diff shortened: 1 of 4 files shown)


rchl:
Now I've realized why the `TSCompletionItem` type existed (it adds extra type information to the `data` object. It would be useful to keep that so I'll look at that part more.

rchl:
> Now I've realized why the `TSCompletionItem` type existed (it adds extra type information to the `data` object. It would be useful to keep that so I'll look at that part more.

Well, it's fine for `data` to be untyped in most cases. Still using `TSCompletionItem` type in `asCompletionItem` where `data` is actually set.

DonnieWest (approved):
LGTM

Issue #217, opened 2021-06-06
Unmaintained dev dependency of a dependency causes tsserver errors and prevents start


dklymenk:
Hello,

I am working on https://github.com/Expensify/Expensify.cash project that has an issue with this language server. It just plain doesn't start. To reproduce, just clone the project, run `npm i` and open a project in your editor.

Here is a relevant log entry:
```
[ ERROR ] 2021-06-06T23:41:13+0300 ] /usr/share/nvim/runtime/lua/vim/lsp/handlers.lua:410 ]     "[tsserver] internal/modules/cjs/loader.js:883\n  throw err;\n
 ^\n\nError: Cannot find module '_debugger'\nRequire stack:\n- /home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver.js\n- /home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver\n    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)\n    at Function.Module._load (internal/modules/cjs/loader.js:725:27)\n    at Module.require (internal/modules/cjs/loader.js:952:19)\n    at require (internal/modules/cjs/helpers.js:88:18)\n    at /home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver.js:39473:25\n    at /home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver.js:39689:7\n    at Object.<anonymous> (/home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver.js:39690:3)\n    at Module._compile (internal/modules/cjs/loader.js:1063:30)\n    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)\n    at Module.load (internal/modules/cjs/loader.js:928:32) {\n  code: 'MODULE_NOT_FOUND',\n  requireStack: [\n    '/home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver.js',\n    '/home/dklymenk/Documents/upwork/Expensify.cash/node_modules/tsc/bin/tsserver'\n  ]\n}\n"
 ```
 
[...]

PR #234, opened 2021-07-07
Lookup tsserver using direct path rather than through .bin alias

rchl:
Lookup the tsserver script through a more explicit typescript/lib/tsserver.js
path rather than through the alias created in the ".bin" directory. This
in theory is less future proof as the path to tsserver could change but
in practice it likely never will. And we also had other code paths
(yarn sdk lookup or lookup within this server's package) that used that
approach already so this unifies things more.

The side-effect of doing that is that we'll use "child_process.fork"
rather than "child_process.spawn" for starting the tsserver but we've
done that in other cases already and it seems like a better approach
anyway as then ENV will be inherited from the parent process (haven't
really verified that).

Also it fixes a corner case like #217 where "tsc" package is used and
makes the server fail.

Relates to #217

### DIFF
--- a/server/src/lsp-server.ts
+++ b/server/src/lsp-server.ts
@@ -69,13 +69,14 @@ export class LspServer {
         if (this.options.tsserverPath) {
             return this.options.tsserverPath;
         }
+        const tsServerPath = path.join('typescript', 'lib', 'tsserver.js');
         // 1) look into .yarn/sdks of workspace root
-        const sdk = findPathToYarnSdk(this.rootPath(), path.join('typescript', 'lib', 'tsserver.js'));
+        const sdk = findPathToYarnSdk(this.rootPath(), tsServerPath);
         if (sdk) {
             return sdk;
         }
         // 2) look into node_modules of workspace root
-        const executable = findPathToModule(this.rootPath(), `.bin/${getTsserverExecutable()}`);
+        const executable = findPathToModule(this.rootPath(), tsServerPath);
         if (executable) {
             return executable;
         }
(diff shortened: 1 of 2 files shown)


rchl:
CC @apexskier

apexskier (approved):
This is similar to the logic I use in [nova-typescript](https://github.com/apexskier/nova-typescript/blob/main/src/tsLibPath.ts#L41-L51) - I use the `tsserver.js` file within the standard node_modules location.

> The side-effect of doing that is that we'll use "child_process.fork" rather than "child_process.spawn" for starting the tsserver but we've done that in other cases already and it seems like a better approach anyway as then ENV will be inherited from the parent process (haven't really verified that).

One issue I've run into with this project when using `child_process.fork` is that it messes with debugging. If I start the `typescript-language-server` binary using `node --inspect-brk` it will bind to port 9229 for debugging, but the forked process _also_ tries to bind to port 9229 and emits warnings. Functionality still works, but it's not ideal.

TypeScript Extension – applyLSPEdits ts - Nova - 2021-07-09 at 17 59 00

Given that the logic still allows [manually specifying a module](https://github.com/theia-ide/typescript-language-server/blob/1005f0d9ea0e79cdf3d79581435c30abae7ec25b/server/src/lsp-server.ts#L69-L71) and [using `spawn`](https://github.com/theia-ide/typescript-language-server/blob/aea14ef2fddb56d13e245b185d496e3de3a440b1/server/src/tsp-client.ts#L106-L109) this should be fine.

> This in theory is less future proof as the path to tsserver could change but in practice it likely never will.

I'm on the same page here.

rchl:
> One issue I've run into with this project when using `child_process.fork` is that it messes with debugging. If I start the `typescript-language-server` binary using `node --inspect-brk` it will bind to port 9229 for debugging, but the forked process _also_ tries to bind to port 9229 and emits warnings. Functionality still works, but it's not ideal.

Thanks for pointing out this issue. I believe I found a solution for this by explicitly setting `execArgv` and thus not inheriting arguments from `process.execArgv`

PR #245, opened 2021-08-08
Chores - Add Credit, Discord link, Remove Lerna, add Github publish workflow

DonnieWest:
I'd appreciate a quick sanity check here:

1. Credit TypeFox for the project
2. Add Discord Links
3. New Publishing workflow
  a. Rip out Lerna
    i. Removes the Lerna based publishing flow, which automatically pushes git tags and other things we don't want
    ii. Forces us to remove the examples directory
    iii. Allows us to use base yarn publishing flow without wrappers
  b. Add simple Github publish workflow
    i. Pushes to NPM when we cut a new tag

If we'd prefer a different approach, just let me know and I'll make modifications before I get too deep here :)


tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
     "files": [],
-    "references": [
-        { "path": "server" },
-        { "path": "example" }
+    "extends": "./tsconfig.settings.json",
```
rchl, line 3: I'd just merge those files now since there is no need to reuse the settings in two places anymore.

rchl:
I think we should also have CHANGELOG.md and automatic script-based creation of releases on Github but I can work on that next. I have experience on that from other projects.

rchl (approved):

rchl:
The `.gitpod.yml` and `.github/gitpod.yml` could also be removed as those likely don't work anymore (the first one refers to script that was removed).

DonnieWest:
Thought I could test this locally without the push-merge dance. Will wrap this up with an alternative testing strategy tomorrow

PR #248, opened 2021-08-15
Add Windows and Mac CI runner

rchl:


rchl:
Will continue after #249 is integrated.
There are some test failures that need to be fixed for Windows.

PR #249, opened 2021-08-15
Fix Windows path regression introduced in #220

rchl:
The Windows paths like "C:\\foo" will not cause "URI.parse" to throw
but instead will create an invalid URI. Change to code to only use
URI.parse for zipfile: paths specifically.


rchl:
Sorry, will merge without additional review as it's a critical issue that breaks Windows.

PR #259, opened 2021-09-18
Implement inlay hints

### LINKED ISSUE #258: Support inlay hints
It would be nice if the inlay hints feature of Typescript 4.4 was exposed through this LSP: https://github.com/microsoft/TypeScript/pull/42089

Or is there already a way and are we wrong to assume (https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils/issues/40) there first needs to be work in this repo?

entropitor:
Fixes #258 

Support inlay hints feature from Typescript 4.4


rchl:
I have fixed the crash in test-utils on master so if you merge latest version, it should work better.

That said, I would like to ask you to add some tests for the feature, if possible.

entropitor:
@rchl What kind of test do you expect? What should it test? That we call tsserver with the correct arguments?

rchl:
Something simple like https://github.com/typescript-language-server/typescript-language-server/blob/413527d172c03d496ca243c72907c9d0064d25cb/src/tsp-client.spec.ts#L53-L66

We make the new request on a certain document and verify that the expected inline hints are returned.

entropitor:
@rchl I've pushed a test

src/tsp-client.spec.ts
```diff
@@ -65,6 +65,32 @@ for (const [serverName, server] of Object.entries({ executableServer, moduleServ
...
+            const inlayHints = await server.request(
+                CommandTypes.ProvideInlayHints,
+                {
+                    file: f,
+                    start: 0,
+                    length: 1000
+                }
+            );
+            // eslint-disable-next-line no-console
+            console.log(inlayHints);
```
rchl, line 89: Please don't leave console logs in the test.

src/tsp-client.spec.ts
```diff
@@ -65,6 +65,32 @@ for (const [serverName, server] of Object.entries({ executableServer, moduleServ
...
+            assert.isDefined(inlayHints.body);
+            assert.equal(inlayHints.body![0].text, ': boolean');
+        }).timeout(10000);
```
rchl, line 92: Can you verify also that all other properties exist?
We want to test that all expected properties are there at least, maybe not necessairly expect specific values since that could in theory change with different TS versions.

src/lsp-extensions.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    export type InlayHintsResult = {
+        inlayHints: InlayHint[];
+    };
+
+    export const type = new lsp.RequestType<
+    InlayHintsParams,
+    InlayHintsResult,
+    lsp.TextDocumentRegistrationOptions
+    >('typescript/inlayHints');
```
rchl, line 32: Have you implemented this based on existing implementation or some specific spec? I know there is attempt to put that feature into the official spec (https://github.com/microsoft/language-server-protocol/pull/1249/files) but the one you have here is not that.

I'm fine with implementing this in the meantime since it uses a custom request. That's not a problem. The only thing I wonder is why is the result an object with `inlayHints` property on it. The result should likely itself just be an array of `InlineHint`.

src/lsp-extensions.ts
```diff
@@ -0,0 +1,39 @@
+import * as lsp from 'vscode-languageserver/node';
```
rchl, line 1: We have a `lsp-protocol.calls.proposed.ts` file for the propesed `textDocument/calls` implementation so I think it would make sense to follow that naming convention and call this one `lsp-protocol.inlayHints.proposed.ts` instead of `lsp-extensions.ts`.

src/tsp-client.spec.ts
```diff
@@ -65,6 +65,32 @@ for (const [serverName, server] of Object.entries({ executableServer, moduleServ
...
+                file: f,
+                fileContent: readContents(f)
+            });
+            await server.request(CommandTypes.Configure, {
+                preferences: {
+                    // @ts-expect-error preference exist
+                    includeInlayFunctionLikeReturnTypeHints: true
+                }
+            });
+            const inlayHints = await server.request(
```
rchl, line 80: I'm sorry if I lead you the wrong path but those tests should actually be in `lsp-server.spec.ts` because we want to test our own API and not TS' one.

src/lsp-extensions.ts
entropitor: I just used something that is more or less equal to what typescript provides as a custom request. We can later migrate to the official spec when it's fully defined.

As for the return type, in general, it's bad practice to return arrays because it will be impossible to return more than the inlayHints if we would ever want that. Now we can just add an extra field.

I can always change if you think it's better to return an array.

entropitor:
@rchl I've pushed some changes

src/lsp-extensions.ts
rchl: > As for the return type, in general, it's bad practice to return arrays because it will be impossible to return more than the inlayHints if we would ever want that. Now we can just add an extra field.

OK. It's an experimental API anyway so it doesn't really matter long-term.

src/lsp-server.spec.ts
```diff
@@ -934,3 +934,41 @@ describe('diagnostics (no client support)', () => {
...
+      `
+        };
+        server.initialize({
+            initializationOptions: {
+                preferences: {
+                    includeInlayFunctionLikeReturnTypeHints: true
+                }
+            },
+            processId: null,
+            capabilities: {},
```
rchl, line 957:
```suggestion
            capabilities: getDefaultClientCapabilities(),
```

src/lsp-server.spec.ts
```diff
@@ -934,3 +934,41 @@ describe('diagnostics (no client support)', () => {
...
+            text: `
+        export function foo() {
+          return 3
+        }
+      `
+        };
+        server.initialize({
+            initializationOptions: {
+                preferences: {
+                    includeInlayFunctionLikeReturnTypeHints: true
```
rchl, line 953: This seems a bit sketchy  because that option doesn't seem to be part of the Typescript `UserPreferences` (https://github.com/microsoft/TypeScript/blob/59fb3731e3858743bd0aae82e35097ab4f4cc7df/src/server/protocol.ts#L3356-L3398)

Also the fact that you have to ignore the type issue in https://github.com/typescript-language-server/typescript-language-server/blob/ca191d2c4091a662dfffbf120e380eec61badcc7/src/tsp-client.spec.ts#L76-L76

Is that because Typescript exposed it in a sloppy way or there is a different way of setting that?

src/lsp-server.ts
```diff
@@ -983,4 +984,47 @@ export class LspServer {
...
+        });
+        const result = await this.tspClient.request(
+            CommandTypes.ProvideInlayHints,
+            {
+                file,
+                start: start,
+                length: end - start
+            }
+        );
+
```
rchl, line 1018: I think we have to cater for the case when the tsserver doesn't support that command since this language server can resolve typescript from the project and there is no guarantee that this will be supported. We could potentially just return an empty result in that case.

rchl (changes requested):

src/lsp-server.spec.ts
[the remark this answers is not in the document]
rchl: I've made this comment before realizing that this is not part of `UserPreferences`. So I retract this issue, at least until I know more what is happening here.

entropitor: The latter is true, the test fails without that line but you cannot set it because it's in something else, not fully UserPreferences. I'm not sure why typescript made it so ugly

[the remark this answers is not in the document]
entropitor: Typescript 4.4 added a bunch of UserPreferences but they are not fully in the type UserPreferences: https://github.com/microsoft/TypeScript/pull/42089/files#diff-9717eb6d935579eea3b5662088dd258d4e1e17e2a405c3f1fa431e981e01b08fR575-R583

I've added the options to the README

entropitor:
@rchl Once more, I think I addressed all comments :)

rchl:
Let me know if you are fine with my minor changes.

rchl:
Thanks!

entropitor:
@rchl can we get this released on npm? Or would we need more PRs to be merged for that?

rchl:
Done.

PR #267, opened 2021-10-10
Fix broken logging

rchl:
The logging didn't work due to my earlier refactor that switched
args from "--foo value" to "--foo=value".


rchl:
No, those places will work either way since the first one is handled by node and the second by `commander`.

PR #270, opened 2021-10-17
Handle the APPLY_COMPLETION_CODE_ACTION command internally

rchl:
Until now, the APPLY_COMPLETION_CODE_ACTION had to be implemented by
each client separately.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -110,7 +110,8 @@ module.exports = {
                 '@typescript-eslint/no-unnecessary-qualifier': 'error',
                 '@typescript-eslint/no-unused-vars': [
                     'error', {
-                        argsIgnorePattern: '^_'
+                        argsIgnorePattern: '^_',
+                        varsIgnorePattern: '^_'
                     }
                 ],
                 '@typescript-eslint/no-useless-constructor': 'error',
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -11,7 +11,7 @@ export const Commands = {
     APPLY_REFACTORING: '_typescript.applyRefactoring',
     ORGANIZE_IMPORTS: '_typescript.organizeImports',
     APPLY_RENAME_FILE: '_typescript.applyRenameFile',
-    /** Commands below should be implemented by the client */
     APPLY_COMPLETION_CODE_ACTION: '_typescript.applyCompletionCodeAction',
+    /** Commands below should be implemented by the client */
     SELECT_REFACTORING: '_typescript.selectRefactoring'
 };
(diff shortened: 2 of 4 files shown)


rchl:
@predragnikolic

predragnikolic:
Just to see if I understand the current behavior.
When resolving a completion item, `asCodeActions` is called:

https://github.com/typescript-language-server/typescript-language-server/blob/dc4704a6ba6e662515ea9edc4885c06d5c6a2279/src/completion.ts#L158

which adds the `APPLY_COMPLETION_CODE_ACTION` command to the completion item.

Currently the client needs to handle the `APPLY_COMPLETION_CODE_ACTION` command when executed.

But after this PR, the server will handle it.
Which sounds ok to me.

-----

BTW On first impression I got confused by this comment:
```
    /** Commands below should be implemented by the client */
    export const APPLY_COMPLETION_CODE_ACTION = "_typescript.applyCompletionCodeAction";
```

And I was looking for a reason why this would need to be implemented by the client.
But given the git history, this was added at the in the first year when the project started,
maybe they wanted to shift some responsibility to the client... but I am not sure.

---
 Btw, I would rename the `asCodeActions` differently, like `assignCommandAndAdditionalTextEdits`, or ideally split the logic to two different functions, `assignCommand` and `assignAdditionalTextEdits` .

`Object.assign(item, assignCommand(...),  assignAdditionalTextEdits(...)); `

predragnikolic:
Can you share your steps on how to test this?

rchl:
> Btw, I would rename the `asCodeActions` differently, like `assignCommandAndAdditionalTextEdits`, or ideally split the logic to two different functions, `assignCommand` and `assignAdditionalTextEdits

I've split it into two functions.

rchl:
@predragnikolic do you want to do a check before I merge it?

predragnikolic:
Unfortunately, I am not able to test it locally at the moment... or in the following week, 
So I would not wait on me :)

🚀 

Everything looks good to me.

But I would put a big note in the CHANGELOG.md 
that clients need to remove the custom handling.
https://github.com/sublimelsp/LSP-typescript/blob/ffc3243681a9f6a8a2fdf6080ee3d1f1869d90a0/plugin.py#L124-L129
to avoid applying the same command two times.

rchl:
> But I would put a big note in the CHANGELOG.md
> that clients need to remove the custom handling.
> [sublimelsp/LSP-typescript@`ffc3243`/plugin.py#L124-L129](https://github.com/sublimelsp/LSP-typescript/blob/ffc3243681a9f6a8a2fdf6080ee3d1f1869d90a0/plugin.py#L124-L129)
> to avoid applying the same command two times.

I think that clients that handle this command now do it by hooking it and thus won't ask server to handle it. So just a note in the changelog saying that the command doesn't need to be handled manually anymore should be enough IMO.

PR #272, opened 2021-10-18
feat: ability to ignore diagnostics by code

### LINKED ISSUE #166: how to disable this may be converted to async function error
I am happy using promises instead of async functions at certain places but typescript language server constantly reports errors saying those functions can be converted to async functions which should be a code action i believe not an error message 
Is there any way to disable certain messages?
Something similar to the `#[allow]` annotation in rust which disables certain warnings

rchl:
Fixes #166

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,10 @@ Some of the preferences can be controlled through the `workspace/didChangeConfig
 [language].inlayHints.includeInlayParameterNameHintsWhenArgumentMatchesName: boolean;
 [language].inlayHints.includeInlayPropertyDeclarationTypeHints: boolean;
 [language].inlayHints.includeInlayVariableTypeHints: boolean;
+// Diagnostics code to be omitted when reporting diagnostics.
+// See https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json for a full list of valid codes.
+diagnostics.ignoredCodes: number[];
+
 ```
 
 # Supported Protocol features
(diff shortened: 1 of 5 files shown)


rchl:
Valid point. I guess I will drop it unless someone else some good arguments for having it.

rchl:
I'll go with that after all. Asking all clients to support that takes quite some effort and convincing and it's not even clear if client authors would want to implement that.

And fixing this will close a bunch of issues.

predragnikolic:
> It's not even clear if client authors would want to implement that.

Yes, it is true that it is not specified in the spec.

Clients already implemented the ability to filter out diagnostics by severity to only show errors and warnings, and hide info, and hints.
That is also not defined in the spec, but yet it made sense for the client to do it.

I do not have anything again this PR,
if this fixes a lot of issues now, why not merge it? :)

rchl:
> Yes, it is true that it is not specified in the spec.

I did not make such argument. :)

> if this fixes a lot of issues now, why not merge it? :)

Well, I guess only this one https://github.com/typescript-language-server/typescript-language-server/issues/142 :P

But still, it's low effort, hopefully no maintenance feature so I don't mind.

drahnieR:
Quite a lot people is suffering from https://github.com/microsoft/vscode/issues/47299 I think

Since that thread is locked, can I link it here to give some hits for folks over there?

predragnikolic:
@jacobchrismarsh coc-tsserver doesn't use this package.
Instead it has ots own implementation for communicating with the tsserver.

jacobchrismarsh:
> @jacobchrismarsh coc-tsserver doesn't use this package. Instead it has ots own implementation for communicating with the tsserver.

Ah, that makes sense. Alright then! Thanks, for the info @predragnikolic

PR #275, opened 2021-10-30
fix: normalize client and tsserver paths

rchl:
This fixes issues with mixed paths on Windows that would result
in server file paths not matching the client file paths and some
functionality to misfunction.

The client can pass URIs in various formats:
 - file://c:/folder/file.ts
 - file://C:/folder/file.ts
 - file://c%3A/folder/file.ts

So both the case of the drive letter can differ and the colon might or
might not be encoded.

As for the file paths returned from the tsserver there can be:
 - C:/folder/file.ts
 - c:/folder/file.ts
 - C:\folder\file.ts

So both the drive letter case and the path separator can differ.

When we are getting URI from the client, we are converting it to a file
path and storing in a map. When the server returns a file path in
some API result then that path might not match what we've gotten from
the client. (Note that the tsserver can auto-discover files itself and
that's where the different path formats are coming from.)

Fix by normalizing both the client paths before storing them in the map
and the server paths before returning to the client or comparing with
client paths.

BTW. The difference in colon encoding should not really be an issue
since this is already normalized when converting from an URI.

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -9,7 +9,7 @@ import * as lsp from 'vscode-languageserver/node';
 import type tsp from 'typescript/lib/protocol';
 import { LspDocument } from './document';
 import { ScriptElementKind } from './tsp-command-types';
-import { asRange, toTextEdit, asPlainText, asDocumentation } from './protocol-translation';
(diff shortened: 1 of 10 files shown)


DonnieWest (approved):
Snagged a coworker to test this on Sublime on Windows. Works well! Thanks

PR #278, opened 2021-11-03
fix: remove unsupported --node-ipc and --socket options

### LINKED ISSUE #111: CLI --socket appears to do nothing
I'm trying to use this tool to do some javascript/typescript analysis/processing and I cannot get a socketed server to run.

if I enter `typescript-language-server --socket 3000` or `node server/lib/cli.js --socket 3000` it just does nothing. No errors, no warnings, no silent process forks... nothing. Note: `--stdio` does seem to create a server, but that is also not particularly useful for my usecase (at present)

what am I doing wrong?

NOTE: i do have globally available tsservers and i've also tried building directly with this repo

rchl:
Fixes #111

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,8 @@ typescript-language-server --stdio
   Options:
 
     -V, --version                          output the version number
-    --stdio                                use stdio
-    --node-ipc                             use node-ipc
+    --stdio                                use stdio (the only supported and required option)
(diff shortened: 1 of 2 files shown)


DonnieWest (approved):
Nice! This should reduce some confusion in the project

Note: I think this was originally done in accordance with https://microsoft.github.io/language-server-protocol/specifications/specification-current/#implementationConsiderations where it's recommended that we support ALL options

But we don't. So not faking it is the correct approach

PR #290, opened 2021-11-12
feat: implement semantic tokens support

thecodrr:
This commit adds support for semantic tokens highlighting. The code should speak for itself — I had transform the output given by the `ts-server`. Supports `full` & `range` semantic tokens.

Tested on Kate Text Editor


src/lsp-server.ts
```diff
@@ -1123,4 +1155,49 @@ export class LspServer {
             ...workspacePreference.inlayHints ?? {}
         };
     }
+
+    async semanticTokens(params: lspsemanticTokens.SemanticTokensParams): Promise<lspsemanticTokens.SemanticTokensResult> {
```
rchl, line 1159: Can we split this into two separate methods and use types directly from the `lsp` import instead of redefining our own?

rchl:
Can you post a screenshot how this part of your code looks like in Kate? I've tried it with ST (special branch that implements semantic token) and saw something that is not ideal (a lot of duplicated colors that make things less readable).

Without semantic tokens:

Screenshot 2021-11-12 at 23 04 12

With semantic tokens:

Screenshot 2021-11-12 at 23 03 48

thecodrr:
@rchl Here are the screenshots. Do note that I am using a development version of Kate so the results might differ:

Without semantic highlighting:

image

With semantic highlighting:

image

> saw something that is not ideal (a lot of duplicated colors that make things less readable).

LSP server shouldn't be concerned about this because it is the client that decides what it wants to do with the tokens. For example, the semantic highlighting in Kate (see above screenshots) is quite selective; for example, it doesn't highlight the variables.

rchl:
Thanks. That looks like an issue in ST's experimental semantic tokens implementation then.

src/lsp-server.ts
```diff
@@ -1123,4 +1155,71 @@ export class LspServer {
...
+            return { data: [], resultId: '' };
+        }
+
+        const start = doc.offsetAt({
+            line: 0,
+            character: 0
+        });
+        const end = doc.offsetAt({
+            line: doc.lineCount,
+            character: 0
```
rchl, line 1177: Does this range goes beneath the actual `end` to simplify things? If so then I guess it's not entirely correct but if tsserver handles that correctly then probably we can keep it.

Could you create at least one simple test that verifies that range request covers whole document, including the the whole last line?

rchl: Never mind, it looks like this will goes past the actual end of the document potentially but if tsserver handles that fine then I don't mind.

rchl: Actually the spec even says:

>  If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line.

so it's all good.

src/lsp-server.ts
```diff
@@ -1123,4 +1155,71 @@ export class LspServer {
             ...workspacePreference.inlayHints ?? {}
         };
     }
+
+    async semanticTokensFull(params: lsp.SemanticTokensParams): Promise<lsp.SemanticTokens> {
+        const file = uriToPath(params.textDocument.uri);
+        this.logger.log('semanticTokensFull', params, file);
+        if (!file) {
+            return { data: [], resultId: '' };
```
rchl, line 1163: Can you remove the `resultId` from results? We don't support deltas and according to the spec `resultId` is optional.

rchl: I've applied that minor change + a removed a couple of unneeded type annotations. The changes look good to me. Just gonna wait for you to give an OK to my changes also.

rchl:
I'll go ahead with it even though you haven't gave a final OK.

Thanks for working on this!

PR #303, opened 2021-11-18
feat: add support for snippet completions of method

### LINKED ISSUE #198: Support function arguments in completion snippets
I came across https://github.com/theia-ide/typescript-language-server/issues/15 while doing some research , and this was closed with a recommendation to OP to make a new issue when it came to snippet support (i.e function parentheses and function arguments / signature) on completion, but never did.

So I thought I'd open one as I would love to see support for this feature.

rchl:
On resolving a completion item, additional information are pulled and the `insertText` of the completion is updated with an extended snippet.

This unfortunately is not supported by the editor I'm using (ST) due to it being unable to resolve the completion before triggering the completion but it will hopefully work in other editors.

Fixes #198

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -8,12 +8,21 @@
 import * as lsp from 'vscode-languageserver/node';
 import type tsp from 'typescript/lib/protocol';
 import { LspDocument } from './document';
-import { KindModifiers, ScriptElementKind } from './tsp-command-types';
+import { CommandTypes, KindModifiers, ScriptElementKind } from './tsp-command-types';
 import { asRange, toTextEdit, asPlainText, asDocumentation, normalizePath } from './protocol-translation';
 import { Commands } from './commands';
+import { TspClient } from './tsp-client';
+import { CompletionOptions, DisplayPartKind } from './ts-protocol';
+import SnippetString from './utils/SnippetString';
+import * as typeConverters from './utils/typeConverters';
 
 interface TSCompletionItem extends lsp.CompletionItem {
     data: tsp.CompletionDetailsRequestArgs;
+}
+
+interface ParameterListParts {
+    readonly parts: ReadonlyArray<tsp.SymbolDisplayPart>;
+    readonly hasOptionalParameters: boolean;
 }
 
 export function asCompletionItem(entry: tsp.CompletionEntry, file: string, position: lsp.Position, document: LspDocument): TSCompletionItem {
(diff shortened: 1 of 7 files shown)


predragnikolic:
Typescript 4.5 added snippets for methods https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#subclass-method-snippets

Is this PR necessary?

rchl:
> Typescript 4.5 added snippets for methods [devblogs.microsoft.com/typescript/announcing-typescript-4-5/#subclass-method-snippets](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#subclass-method-snippets)
> 
> Is this PR necessary?

Yes. Those are unrelated features. The above-mentioned is for class members specifically and is provided in initial completion already, without need of resolving.

src/completion.ts
```diff
@@ -177,17 +186,135 @@ function asCommitCharacters(kind: ScriptElementKind): string[] | undefined {
...
+    const parameterListParts = getParameterListParts(displayParts);
+    const snippet = new SnippetString();
+    snippet.appendText(`${item.insertText || item.label}(`);
+    appendJoinedPlaceholders(snippet, parameterListParts.parts, ', ');
+    if (parameterListParts.hasOptionalParameters) {
+        snippet.appendTabstop();
+    }
+    snippet.appendText(')');
+    snippet.appendTabstop(0);
+    item.insertText = snippet.value;
```
yyoncho, line 248: `insertText` should not be modified as part of the resolve request but it should be populated in advance as part of the textDocument/completion request.

Refer to https://microsoft.github.io/language-server-protocol/specification#textDocument_completion . If you do so, you are forcing the client to do blocking call to the server before inserting the text.

rchl: From the link you've provided:

> servers can additionally provide a handler for the completion item resolve request (‘completionItem/resolve’). This request is sent when a completion item is selected in the user interface.

Resolving all the completions up-front (and there could be thousands of them) would be insanely slow.

rchl: I see what you mean now... You mean that `completionItem/resolve` should not be used and everything should be resolved up-front. It might be an option for some servers but not here because resolving requires making an additional request to `tsserver` and with thousands of completions present (or even when completions are in tens or hundreds) is just not a viable option.

yyoncho:
```
All other properties (usually sortText, filterText, insertText and textEdit) must be provided in the [textDocument/completion](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) response and must not be changed during resolve.
```
(note the list of the properties that have to be prepopulated). Note also, that the request to the server when the item is selected might be async. We had issues like that in the past and that is why I asked MS team to explicitly specify what can and what cannot be missing in the initial request. As for the server side performance: you can return partial results if calculating the snippet is expensive which for the most of the server is not the case because the methods are usually obtained with their signatures.

yyoncho: > an additional request to `tsserver` and with thousands of completions present (or even when completions are in tens or hundreds) is just not a viable option.

That is why you can return partial results in textDocument/completion. Either way, returning thousands of results is not a good idea.

rchl: My understanding of that specification is different. It says:

> Since 3.16.0 the client can signal that it can resolve more properties lazily. This is done using the completionItem#resolveSupport client capability which lists all properties that can be filled in during a ‘completionItem/resolve’ request. All other properties (usually sortText, filterText, insertText and textEdit) must be provided in the [textDocument/completion](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) response and must not be changed during resolve.

Note how it says "usually" which is not the same as "must not".
In my understanding the "All other properties" sentence is relative to the what was specified in `completionItem#resolveSupport` and there are no restrictions on what can be specified there (AFAICS).

But if my understanding is not correct and the expectation is that the insertText or textEdit are not changed then it will be effectively impossible to provide support for this feature then.

> As for the server side performance: you can return partial results if calculating the snippet is expensive

In that case it is. Not sure what are you suggesting here though. The results are already partial without resolving. I suppose there could be an option to disable resolving

rchl: > Either way, returning thousands of results is not a good idea.

Separate discussion but there is no way reasonably decide what to limit what the tsserver returns.

yyoncho: FWIW the reference lsp client vscode works in a similar manner - it inserts the data and then calls resolve to obtain additionalTextEdits - https://github.com/microsoft/language-server-protocol/issues/1003#issuecomment-653499069

yyoncho: > Sure but it doesn't do that for the Typescript language services as it is not running those through LSP so it's a bit irrelevant in this case.

My point is that other clients might have similar implementation and most likely will suffer from the same issue.

FWIW LSP protocol also supports including commands in CompletionItem so that implementation can work.

yyoncho: I meant that it can apply the necessary edits to the buffer and expand the snippet as a client-side command. Not sure if this will work well.

rchl: Then the implementation of such custom logic would need to be duplicated for each client.

And also it would be non-trivial at best to revert the completion that was inserted and generate a correct location for the new snippet. Or whatever the approach would have to be...

I'm personally not gonna attempt that myself as there are multiple things I don't like about that approach already.

wyuenho: I'm late to the party so pardon me. I see on the latest, you already [support textEdit](https://github.com/typescript-language-server/typescript-language-server/blob/master/src/completion.ts#L422-L424), perhaps we can delete this line?

According to the comment on `CompletionItem#insertText` in the latest version of the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem), it says:

> /**
	 * A string that should be inserted into a document when selecting
	 * this completion. When omitted the label is used as the insert text
	 * for this item.
	 *
	 * The `insertText` is subject to interpretation by the client side.
	 * Some tools might not take the string literally. For example
	 * VS Code when code complete is requested in this example
	 * `con<cursor position>` and a completion item with an `insertText` of
	 * `console` is provided it will only insert `sole`. Therefore it is
	 * recommended to use `textEdit` instead since it avoids additional client
	 * side interpretation.
	 */
	 
There's no need to keep this line anymore and clients are suppose to prefer textEdit when it exists, which is what lsp-mode does now.

rchl: Are you suggesting to remove the `insertText` property?

If so then I'm not sure what would that achieve. Can you specify what is your goal with that?

wyuenho: There are a couple of problems with the existing implementation of `createSnippetOfFunctionCall`.

1. If the client has not requested delay computation of `insertText` by listing it in `resolveSupport`, there is no expectation that it should change in a resolution. (I notice the same issue in [`asResolvedCompletionItem`](https://github.com/typescript-language-server/typescript-language-server/blob/master/src/completion.ts#L350-L378) as well.)
2. The ["snippet" can be just a label](https://github.com/typescript-language-server/typescript-language-server/blob/184c60de3308621380469d6632bdff2e10f672fd/src/completion.ts#L413), or [some prefix + a label](https://github.com/typescript-language-server/typescript-language-server/blob/184c60de3308621380469d6632bdff2e10f672fd/src/completion.ts#L149), in that case there is also no reason to reformat it to a snippet and change the `insertTextFormat` property to snippet instead of leaving it as plaintext.
3. Given that the completion item from the response of `textDocument/completion` [will only contain an `insertText` property when there's no range](https://github.com/typescript-language-server/typescript-language-server/blob/184c60de3308621380469d6632bdff2e10f672fd/src/completion.ts#L201), and you can't construct a snippet until you have a `detail` during resolution due to performance concerns and parity with tsserver, would it not be easier to just give up on dynamically generating snippets during resolution, and just use tsserver's builtin snippets?

There's probably more than one line of changes needed to conform to the spec, but these are my thoughts.

wyuenho: Now that I've looked at it a little more closer, it appears there's a lot more code you can delete, all those filterText, CompletionContext stuff can all be gone.

rchl: Hard to me to reply to that since I don't have this logic fresh in my memory and I don't remember in which cases it applies exactly but I would expect that dropping some of this code would result in dropping some functionality. So we'd have to evaluate what is being lost and whether it's acceptable.

wyuenho: I've taken a third look at the code base, it seems most of this package is an attempt to port VS Code's [typescript-language-features](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features) before it was refactored into its own package, and quite a bit have diverged between the two in the last 6 years. I think I've going to give [vtsls](https://github.com/yioneko/vtsls) and [coc-typescript ](https://github.com/neoclide/coc-tsserver)a try as they seem like much more reasonable approaches to putting an LSP layer on top of typescript-language-features.

rchl: Personal preference. I keep using this one mainly because it works better with Sublime Text due to custom configuration like `useLabelDetailsInCompletionEntries`. I'm not sure sure if vtsls would want to maintain such custom logic due to its goals of only keeping minimal patches on top of the original code.

PR #305, opened 2021-11-20
fix: exit the server if tsserver process crashes

### LINKED ISSUE #302: tsserver process dying is not handled
Sometimes the `tsserver` process dies (OOM or other reasons), when that happens the parent process is essentially useless. 

Happy to contribute here, but would love some advice on how you think we should proceed when the process dies. Some options:
1. Kill the `typescript-language-server` process, probably the easiest option, let the user handle restarting
2. Respawn `tsserver` process internally inside `TspClient`, reconfigure `tsserver` (need to cache initial configuration and compiler options), and then resend pending requests.
3. Create a new instance of `TspClient` inside `LspClient`, ditto above

rchl:
Fixes #302

### DIFF
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -153,6 +153,9 @@ export class LspServer {
         if (!started) {
             throw new Error('tsserver process has failed to start.');
         }
+        process.on('exit', () => {
+            this.tspClient.shutdown();
+        });
         this.tspClient.request(CommandTypes.Configure, {
             ...hostInfo ? { hostInfo } : {},
             formatOptions: {
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -145,8 +145,8 @@ export class LspServer {
         });
 
         const started = this.tspClient.start({
-            onExit: (exitCode) => {
-                this.logger.error(`tsserver process has exited with status code "${exitCode}". Stopping the server.`);
+            onExit: (exitCode, signal) => {
+                this.logger.error(`tsserver process has exited (exit code: ${exitCode}, signal: ${signal}). Stopping the server.`);
                 // Allow the log to be dispatched to the client.
                 setTimeout(() => process.exit(1));
             }
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -141,16 +141,15 @@ export class LspServer {
             globalPlugins,
             pluginProbeLocations,
             logger: this.options.logger,
-            onEvent: this.onTsEvent.bind(this)
-        });
-
-        const started = this.tspClient.start({
+            onEvent: this.onTsEvent.bind(this),
             onExit: (exitCode, signal) => {
                 this.logger.error(`tsserver process has exited (exit code: ${exitCode}, signal: ${signal}). Stopping the server.`);
                 // Allow the log to be dispatched to the client.
                 setTimeout(() => process.exit(1));
             }
         });
+
+        const started = this.tspClient.start();
         if (!started) {
             throw new Error('tsserver process has failed to start.');
         }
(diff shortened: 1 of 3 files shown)


src/tsp-client.ts
```diff
@@ -147,6 +149,13 @@ export class TspClient {
             const stringMsg = typeof data === 'string' ? data : dec.write(data);
             this.tsserverLogger.error(stringMsg);
         });
+        return true;
+    }
+
+    shutdown(): void {
```
masad-frost, line 155: Doesn't look like you ended up using this.

src/lsp-server.ts
```diff
@@ -143,7 +143,16 @@ export class LspServer {
             onEvent: this.onTsEvent.bind(this)
         });
 
-        this.tspClient.start();
+        const started = this.tspClient.start({
+            onExit: (exitCode) => {
+                this.logger.error(`tsserver process has exited with status code "${exitCode}". Stopping the server.`);
+                // Allow the log to be dispatched to the client.
```
masad-frost, line 149: Good call

src/lsp-server.ts
```diff
@@ -143,7 +143,16 @@ export class LspServer {
...
+                setTimeout(() => process.exit(1));
+            }
+        });
+        if (!started) {
+            throw new Error('tsserver process has failed to start.');
```
masad-frost, line 154: What happens after this is thrown? Previously TspClient would exit the process if not started, does this propagate to a handler that will exit?

masad-frost (approved):
Thank you! beat me to it.

Left a couple of comments

src/lsp-server.ts
rchl: I've tested by always throwing and in that case the LSP library is returning an error to the client which makes the client close the process.

src/tsp-client.ts
rchl: Yes. Though I wonder now whether I should restore the `process.on('exit', ...)` handler and call the `shutdown`. As far as my testing shows the child tsserver process is closed automatically but not sure if that's gonna be the case in all situations and on all platforms...

masad-frost: Yeah, it definitely depends on the OS and on Linux's pid1 implementation. I'd say it's safer if you did add it.

rchl:
@masad-frost are you OK with latest changes?

masad-frost:
Yeah, makes sense. I think people also trap SIGINT as it can exit without triggering an exit event, I'm not sure why that is.

rchl:
There is some related information in https://nodejs.org/docs/latest-v14.x/api/child_process.html#child_process_event_exit but to be fair I'm still not sure if I need to handle those.

I've updated the code to also report a signal (in addition to exit code) and tested in those ways:
 - set `maxTsServerMemory` to `50` to trigger OOM - gives `exit code: null, signal: SIGABRT`
 - manually kill tsserver process from activity monitor using `Quit` option - gives `exit code: null, signal: SIGTERM`
 - manually kill tsserver process using `Force quit` option  - gives `exit code: null, signal: SIGKILL`

So it seems like I don't really need to handle those signals.

rchl:
Someone on the SO thread also said:

> "exit" is an event that gets triggered when node finish it's event loop internally, it's not triggered when you terminate the process externally.

But that clearly doesn't seem to be the case here. Maybe because it's a subprocess that we are talking here and not the main process. I don't know.

masad-frost:
Good finds. Makes sense to me.

I meant SIGINT for the main CLI process, for cleaning up the child process, like you did here https://github.com/typescript-language-server/typescript-language-server/pull/305/files#diff-2a5096ffdb93efd5ef847c28e268f1e1e824fa2894ecd611c1a1f1164c4f4c63R157-R159

masad-frost:
:shipit:

PR #314, opened 2021-11-25
fix: lookup workspace typescript in dirs higher up the tree also

### LINKED ISSUE #313: `typescript` version is no longer resolved above the `rootUri`
Thanks for the work maintain this project!

Prior to #292 it looks like ts-ls would resolve the version of `typescript` following node module resolution (up the folder structure). This is useful if you're in workspaces that has `root_dir` like folders but the `typescript` module is installed higher in the tree, something like this

```console
├── index.ts
├── node_modules
│   └── typescript
├── package.json
├── packages
│   └── some_app
│       ├── package.json
│       ├── tsconfig.json
│       └── index.ts        <--- open this file
└── tsconfig.json
```

This setup can be common in yarn / npm workspace setups.

We could reimplement something like the `findPathToYarnSdk` at the point we look for the various typescript module options (`getWorkspaceVersion`).
https://github.com/typescript-language-server/typescript-language-server/blob/42bf57a5af9672275437fa77ef0d4fcd94084d16/src/modules-resolver.ts#L18-L30
https://github.com/typescript-language-server/typescript-language-server/blob/51f64ed1a74bb90fd24837a383826601ff8c1ddb/src/utils/versionProvider.ts#L136-L149

What do you think? I'd be happy to look at picking this up

tapayne88:
Resolves #313

### DIFF
--- a/src/utils/modules-resolver.spec.ts
+++ b/src/utils/modules-resolver.spec.ts
@@ -1,0 +1,16 @@
+import * as chai from 'chai';
+import * as path from 'path';
+import { findPathToModule } from './modules-resolver';
+import { MODULE_FOLDERS } from './versionProvider';
+
+describe('findPathToModule', () => {
+    it('resolves tsserver in own directory', () => {
+        const dir = path.join(__dirname, '../..');
+        const tsserverPath = findPathToModule(dir, MODULE_FOLDERS);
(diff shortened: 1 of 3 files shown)


rchl (approved):

rchl:
Thanks!

tapayne88:
Awesome, thanks for your help @rchl!

PR #318, opened 2021-12-02
feat: implement additional code actions for handling auto-fixing

### LINKED ISSUE #235: Support for auto-import all
I just moved away from coc-tsserver, and I really miss the auto-import all functionality.
I was wondering if this is something that can be included in this server.

rchl:
Implements additional code actions that can be triggered using the Code-Actions-On-Save functionality of the editor or through "Source actions" menu.

 - `source.addMissingImports.ts` - adds imports for used but not imported symbols
 - `source.fixAll.ts` - despite the name, fixes a couple of specific issues: unreachable code, await in non-async functions, incorrectly implemented interface
 - `source.removeUnused.ts` - removes declared but unused variables
 - `source.organizeImports.ts` (was already supported but renamed the suffix from `ts-ls` to `ts`) - organizes and removes unused imports

Those are not exposed in the normal code action menu that is shown for individual diagnostics. This matches the behavior of VSCode but not sure about COC.

Fixes #235

### DIFF
--- a/src/features/fix-all.ts
+++ b/src/features/fix-all.ts
@@ -1,0 +1,206 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import tsp, { CommandTypes } from 'typescript/lib/protocol';
+import * as lsp from 'vscode-languageserver/node';
+import { CodeActions } from '../commands';
+import { LspDocuments } from '../document';
+import { toFileRangeRequestArgs, toTextDocumentEdit } from '../protocol-translation';
+import { TspClient } from '../tsp-client';
+import * as errorCodes from '../utils/errorCodes';
+import * as fixNames from '../utils/fixNames';
+
+interface AutoFix {
+    readonly codes: Set<number>;
+    readonly fixName: string;
+}
+
+async function buildIndividualFixes(
+    fixes: readonly AutoFix[],
+    client: TspClient,
+    file: string,
+    documents: LspDocuments,
+    diagnostics: readonly lsp.Diagnostic[]
+): Promise<lsp.TextDocumentEdit[]> {
+    const edits: lsp.TextDocumentEdit[] = [];
+    for (const diagnostic of diagnostics) {
+        for (const { codes, fixName } of fixes) {
+            if (!codes.has(diagnostic.code as number)) {
+                continue;
+            }
+
+            const args: tsp.CodeFixRequestArgs = {
+                ...toFileRangeRequestArgs(file, diagnostic.range),
+                errorCodes: [+diagnostic.code!]
+            };
+
+            const response = await client.request(CommandTypes.GetCodeFixes, args);
+            if (response.type !== 'response') {
+                continue;
+            }
+
+            const fix = response.body?.find(fix => fix.fixName === fixName);
+            if (fix) {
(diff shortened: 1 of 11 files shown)


src/features/fix-all.ts
```diff
@@ -0,0 +1,206 @@
...
+            const combinedResponse = await client.request(CommandTypes.GetCombinedCodeFix, combinedArgs);
+            if (combinedResponse.type !== 'response' || !combinedResponse.body) {
+                return edits;
+            }
+
+            edits.push(...combinedResponse.body.changes.map(change => toTextDocumentEdit(change, documents)));
+            return edits;
+        }
+    }
+    return edits;
```
predragnikolic, line 105: Just an observation.

The `buildIndividualFixes` and `buildCombinedFix` do a lot of same logic,
wiith the exception that the `buildCombinedFix` has some additional code that fires the `CommandTypes.GetCombinedCodeFix` request, and adds additional text edits... 

The only confusing thing is when is preferable to call `buildIndividualFixes` and when it is preferable to call `buildCombinedFix`?

predragnikolic:
Tested it, 
works as described! :)

src/features/fix-all.ts
rchl: I don't know. This is mostly just code copied from VSCode's typescript language features. They know best what to do since they work on both the language features and typescript.

predragnikolic:
The only thing I noticed is when I have the same symbol name, exported from multiple files.
```ts
// foo.ts
export const hello = "the one that I do NOT want"
```

```ts
// bar.ts
export const hello = "the one that I want"
```

and when I have a file `baz.ts`
with the following content:
```ts
// baz.ts
hello
^^^^^ - not imported
```

When running the `source.addMissingImports.ts`,
I will get something like this:
```ts
//baz
import { hello } from './foo'; // I implicitly expected to import from ./bar
hello
```
Which is maybe not what someone would want to import...

I would expect that `source.addMissingImports.ts` doesn't import anything in this case, 
to avoid issues where you accidentally import something that you do not want.

But the behavior in this PR is also the same behavior in VS Code. 
So maybe that is by design, but I would argue that this case could be handled more nicely.

rchl:
> I would expect that `source.addMissingImports.ts` doesn't import anything in this case,
> to avoid issues where you accidentally import something that you do not want.

That's a pretty good point. I've tried to only provide the fix if there is only one suggestion but the problem then is that there is almost always two suggestions so the code action rarely works then. At least when importing from built-in node modules there is (almost?) always both an import for `foo` and `node:foo`.

So maybe it makes sense to do something like that but the logic would have to be smarter than "if there is only one fix".

rchl:
What's more, TS seems to be providing duplicate suggestions. This is a result for a document with just `existsSync('t');` (the `changes` are the same for the duplicate pairs):

```
[
  {
    fixName: 'import',
    description: `Import 'existsSync' from module "fs"`,
    changes: [ [Object] ]
  },
  {
    fixName: 'import',
    description: `Import 'existsSync' from module "fs"`,
    changes: [ [Object] ]
  },
  {
    fixName: 'import',
    description: `Import 'existsSync' from module "node:fs"`,
    changes: [ [Object] ]
  },
  {
    fixName: 'import',
    description: `Import 'existsSync' from module "node:fs"`,
    changes: [ [Object] ]
  },
  {
    fixName: 'import',
    description: `Import 'existsSync' from module "fs-extra"`,
    changes: [ [Object] ]
  }
]
```

rchl:
Will get it in for now. Can be improved later if someone cares enough.

PR #326, opened 2021-12-07
feat: report progress when loading the project

### LINKED ISSUE #237: Ability to send `$/progress` messages
When working on large repos, it can take ~1min for typescript server from `textDocument/didOpen` ->  `textDocument/publishDiagnostics`.

Is it possible to add the ability to send send `$/progress` messages if the client attaches a progress token to the request so that users are able to convert this to some type of "loading" indicator e.g. in vim's status bar?

https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#initiatingWorkDoneProgress

rchl:
Added server-initiated progress that notifies the client when the typescript project is being loaded. This should trigger on initial loading and also on changing project settings in `tsconfig.json`. Basically any time Typescript decides to (re)load the project since it's based on typescript notifications.

Fixes #237

### DIFF
--- a/src/lsp-client.ts
+++ b/src/lsp-client.ts
@@ -6,9 +6,18 @@
  */
 
 import * as lsp from 'vscode-languageserver/node';
+import { WorkDoneProgressServerReporter } from 'vscode-languageserver/node';
 import { TypeScriptRenameRequest } from './ts-protocol';
 
+export interface ProgressReporter {
+    begin(message?: string): void;
+    report(message: string): void;
+    end(): void;
+}
+
 export interface LspClient {
+    setClientCapabilites(capabilites: lsp.ClientCapabilities): void;
+    createProgressReporter(): ProgressReporter;
     publishDiagnostics(args: lsp.PublishDiagnosticsParams): void;
     showMessage(args: lsp.ShowMessageParams): void;
     logMessage(args: lsp.LogMessageParams): void;
@@ -18,7 +27,47 @@ export interface LspClient {
 }
 
 export class LspClientImpl implements LspClient {
-    constructor(protected connection: lsp.Connection) {
+    private clientCapabilities?: lsp.ClientCapabilities;
+
+    constructor(protected connection: lsp.Connection) {}
+
+    setClientCapabilites(capabilites: lsp.ClientCapabilities): void {
+        this.clientCapabilities = capabilites;
+    }
+
+    createProgressReporter(): ProgressReporter {
+        let workDoneProgress: Promise<WorkDoneProgressServerReporter> | undefined;
+        return {
+            begin: (message = '') => {
+                if (this.clientCapabilities?.window?.workDoneProgress) {
+                    workDoneProgress = this.connection.window.createWorkDoneProgress();
+                    workDoneProgress
+                        .then((progress) => {
+                            progress.begin(message);
+                        })
+                        .catch(() => {});
+                }
+            },
+            report: (message: string) => {
+                if (workDoneProgress) {
+                    workDoneProgress
+                        .then((progress) => {
+                            progress.report(message);
+                        })
+                        .catch(() => {});
+                }
+            },
+            end: () => {
+                if (workDoneProgress) {
+                    workDoneProgress
+                        .then((progress) => {
+                            progress.done();
+                        })
+                        .catch(() => {});
+                    workDoneProgress = undefined;
+                }
+            }
+        };
     }
 
     publishDiagnostics(args: lsp.PublishDiagnosticsParams): void {
--- a/src/lsp-client.ts
+++ b/src/lsp-client.ts
@@ -6,7 +6,6 @@
  */
 
 import * as lsp from 'vscode-languageserver/node';
-import { WorkDoneProgressServerReporter } from 'vscode-languageserver/node';
 import { TypeScriptRenameRequest } from './ts-protocol';
 
 export interface ProgressReporter {
@@ -36,7 +35,7 @@ export class LspClientImpl implements LspClient {
     }
 
     createProgressReporter(): ProgressReporter {
-        let workDoneProgress: Promise<WorkDoneProgressServerReporter> | undefined;
+        let workDoneProgress: Promise<lsp.WorkDoneProgressServerReporter> | undefined;
         return {
             begin: (message = '') => {
                 if (this.clientCapabilities?.window?.workDoneProgress) {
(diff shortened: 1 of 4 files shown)


src/lsp-server.ts
```diff
@@ -38,6 +38,52 @@ import { computeCallers, computeCallees } from './calls';
...
+        this._progressReporter = this.lspClient.createProgressReporter();
+        this._progressReporter.begin('Initializing JS/TS language features');
+
+        new Promise<void>((resolve, reject) => {
+            this._loadingProject = { project: projectName, resolve, reject };
+        });
+    }
+
+    public finishedLoadingProject(projectName: string | undefined): void {
+        if (this._loadingProject && this._loadingProject.project === projectName) {
```
predragnikolic: Hah, I wanted to suggest that you could use the `this._loadingProject?.projec` instead of `(this._loadingProject && this._loadingProject.project === projectName`,
but apparently ts will still complain bellow.

image

rchl: just simplified it since creating a promise was pointless in our case

rchl: but yes, it wouldn't be the same logic when using optional chaining since it could then evaluate to true when `undefined === undefined`

src/lsp-server.ts
```diff
@@ -38,6 +38,47 @@ import { computeCallers, computeCallees } from './calls';
...
+        // TS projects are loaded sequentially. Cancel existing task because it should always be resolved before
+        // the incoming project loading task is.
+        this.reset();
+
+        this._loadingProjectName = projectName;
+        this._progressReporter = this.lspClient.createProgressReporter();
+        this._progressReporter.begin('Initializing JS/TS language features');
+    }
+
+    public finishedLoadingProject(projectName: string | undefined): void {
```
predragnikolic, line 72: can the project type be just a string?
```suggestion
    public finishedLoadingProject(projectName: string): void {
```

src/lsp-server.ts
```diff
@@ -38,6 +38,47 @@ import { computeCallers, computeCallees } from './calls';
...
+        if (this._loadingProjectName && this._loadingProjectName === projectName) {
```
predragnikolic, line 73: can the check be just?
`
```suggestion
        if (this._loadingProjectName === projectName) {
```

rchl: Same answer as in https://github.com/typescript-language-server/typescript-language-server/pull/326#discussion_r764363287

But if I'm to trust Typescript types, I can change the `projectName` argument to `string` and then it will be fine.

src/lsp-server.ts
rchl: If I'm to trust Typescript types then this should be OK.

src/lsp-client.ts
```diff
@@ -18,7 +26,47 @@ export interface LspClient {
...
+        let workDoneProgress: Promise<lsp.WorkDoneProgressServerReporter> | undefined;
+        return {
+            begin: (message = '') => {
+                if (this.clientCapabilities?.window?.workDoneProgress) {
+                    workDoneProgress = this.connection.window.createWorkDoneProgress();
+                    workDoneProgress
+                        .then((progress) => {
+                            progress.begin(message);
+                        })
+                        .catch(() => {});
```
predragnikolic, line 47: can the `createWorkDoneProgress` throw an exception?
is it ok to silence the exception `catch(() => {});` ?

predragnikolic: If it doesn't throw an exception, 
IMO the async/await syntax is more easier to read. 

```ts
            begin: async (message = '') => {
                if (this.clientCapabilities?.window?.workDoneProgress) {
                    const progress = await this.connection.window.createWorkDoneProgress();
                    progress.begin(message);
                }
            },
```

rchl: It technically can reject since the client can refuse to create the token.

This code was inspired by something I found in pyright but personally I think it makes more sense this way.

We don't really need to await those methods but when using `async` function it is sort of expected that those are awaited and I don't want to do that.

src/lsp-client.ts
```diff
@@ -18,7 +26,47 @@ export interface LspClient {
...
+                }
+            },
+            report: (message: string) => {
```
predragnikolic, line 50: Just an observation.
this report is not currently used in the project?

Because the ts service only has `EventTypes.ProjectLoadingStart`, `EventTypes.ProjectLoadingFinish` events. There is no event that can be used for this `report: (message: string)` function. Is that right?

src/lsp-client.ts
predragnikolic: Github has a bug on my end... 
I do not see the resolve conversation button again :)

rchl: Happens to me to sometimes and I can't figure out what's the Github's logic there...

src/lsp-client.ts
rchl: Yeah but I'll just leave it to make the interface future-proof since it's so little extra code.

src/lsp-server.ts
```diff
@@ -38,6 +38,44 @@ import { computeCallers, computeCallees } from './calls';
...
+import { LspClient, ProgressReporter } from './lsp-client';
+
+class ServerInitializingIndicator {
+    private _loadingProjectName?: string;
+    private _progressReporter?: ProgressReporter;
+
+    constructor(private lspClient: LspClient) {}
+
+    public reset(): void {
+        if (this._loadingProjectName) {
```
predragnikolic, line 50: do you like early returns to keep the indentation minimal?

rchl: Since this is based on typescript-language-features, I wasn't really trying to rewrite everything.

Generally I prefer early returns, yes.

rchl: In this case it would result in more lines which is another thing I prefer to avoid though :)

predragnikolic (approved):
Tested.

There is a message when the projects starts 'Initializing JS/TS language features'.
There is no message that reports the progress (like 67% of the project is loaded) as I might have initially expected, which would be useful.
And there is no message at the end when the project has been fully.

IMO, for me this is too much code for my taste just to show a 'Initializing JS/TS language features', but I do not have anything against merging this, as the code is readable.

rchl:
We (or Typescript) doesn't know how long the loading will take so there is no way to show a progress.

Also it's implied that the project is loading when the string is shown so I wouldn't see any benefit in showing extra message when it has loaded.

This is especially useful for large projects that take many seconds to load. Without the message it's not clear why none of the features (completions, type checking) seem to be working.

PR #334, opened 2021-12-13
fix: correct matching of "only" kinds provided by the client

rchl:
The matching of "only" code action kind provided by the client was not correct (especially for the "fixall" actions). Imported the `CodeActionKind` class from VSCode and using its matching logic to fix that.

### DIFF
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -13,7 +13,6 @@ import * as lspinlayHints from './lsp-protocol.inlayHints.proposed';
 import * as lspsemanticTokens from './semantic-tokens';
 import tsp from 'typescript/lib/protocol';
 import * as fs from 'fs-extra';
-import { CodeActionKind, FormattingOptions } from 'vscode-languageserver/node';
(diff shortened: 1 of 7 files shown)


src/lsp-server.ts
```diff
@@ -851,7 +851,7 @@ export class LspServer {
         // pending diagnostic requests (regardless of for which file).
         // In general would be better to replace the whole diagnostics handling logic with the one from
         // bufferSyncSupport.ts in VSCode's typescript language features.
-        if (!this.pendingDebouncedRequest && only?.some(kind => kind === CodeActionKind.Source || kind.startsWith(CodeActionKind.Source + '.'))) {
+        if (!this.pendingDebouncedRequest && kinds?.some(kind => TypeScriptAutoFixProvider.kinds.some(k => k.contains(kind)))) {
```
rchl, line 854: This line specifically was not correct as it has matched TS kinds even when the provided kind was `source.foo`, for example.

rchl: Now fixed.

PR #337, opened 2021-12-18
fix: help users resolve no valid tsserver version error

### LINKED ISSUE #336: Unclear how to resolve "Could not find a valid tsserver version. Exiting." error
When the user receives this error on initialize `Could not find a valid tsserver version. Exiting.` it is not clear what needs to be done to resolve the issue. They may not have any idea what "tsserver" is (and may be confused into thinking that tsserver is typescript language server). Judging from https://github.com/Microsoft/TypeScript/wiki/Standalone-Server-%28tsserver%29 this can usually be resolved by `npm install -g typescript`.

axelson:
Not all users will already have typescript installed (or they may have done something like switching to a new nodejs install via asdf that doesn't have typescript installed). Also since tsserver is a dependency of typescript-language-server I added it to the readme.

Related: https://github.com/typescript-language-server/typescript-language-server/issues/28
Fixes #336

### DIFF
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -174,7 +174,7 @@ export class LspServer {
         if (typescriptVersion) {
             this.logger.info(`Using Typescript version (${typescriptVersion.source}) ${typescriptVersion.versionString} from path "${typescriptVersion.tsServerPath}"`);
         } else {
-            throw Error('Could not find a valid tsserver version on $PATH. Please install tsserver with `npm install -g typescript`. Exiting.');
+            throw Error('Could not find a valid tsserver executable in the workspace or in the $PATH. Please ensure that the "typescript" dependency is installed in either location. Exiting.');
         }
 
         this.tspClient = new TspClient({
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -174,7 +174,7 @@ export class LspServer {
         if (typescriptVersion) {
             this.logger.info(`Using Typescript version (${typescriptVersion.source}) ${typescriptVersion.versionString} from path "${typescriptVersion.tsServerPath}"`);
         } else {
-            throw Error('Could not find a valid tsserver version. Exiting.');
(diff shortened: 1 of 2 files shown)


src/lsp-server.ts
```diff
@@ -174,7 +174,7 @@ export class LspServer {
         if (typescriptVersion) {
             this.logger.info(`Using Typescript version (${typescriptVersion.source}) ${typescriptVersion.versionString} from path "${typescriptVersion.tsServerPath}"`);
         } else {
-            throw Error('Could not find a valid tsserver version. Exiting.');
+            throw Error('Could not find a valid tsserver version on $PATH. Please install tsserver with `npm install -g typescript`. Exiting.');
```
rchl, line 177: `findTypescriptVersion` will try to find the typescript dependency either in the workspace or globally so assuming that typescript was not found in the PATH and suggesting to install it globally is only half-correct.

Also some users might be using `yarn` or other package manager so maybe better leave out the exact command.

I suppose the message could mention both possibilities like:

```suggestion
            throw Error('Could not find a valid tsserver version in the workspace or in the $PATH. Please ensure that the "typescript" dependency is installed in either location. Exiting.');
```

rchl (changes requested):

src/lsp-server.ts
axelson: Ah, that's a good point. What about

> Could not find a valid `tsserver` executable in the workspace or in the $PATH. Please ensure that the "typescript" dependency is installed in either location. Exiting.

That way it is more clear about what binary executable needs to be available.

rchl: Sounds good.

David-Else:
Plz add this, it keeps on catching me out!

src/lsp-server.ts
rchl: Are you planning to continue here or you want me to take over?

axelson: Ah, I didn't realize this was outstanding on my part. Updated to the proposed wording in 7f662ba92ddf5f40935e02f24310e5e430802e25

rchl (approved):

PR #348, opened 2022-01-01
fix: pass format options for organizing import

yioneko:
Related: https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils/discussions/95

`organizeImport` should respect the formatting options.


rchl:
There is another place where organizeImport is called - https://github.com/typescript-language-server/typescript-language-server/blob/dc4266d0d0cfa971fb794cfe4af99d0a747283ff/src/lsp-server.ts#L926-L926

Can you also configure before that?

PR #361, opened 2022-01-05
feat: workspace diagnostics poc

entropitor:
Proof of concept of the workspace diagnostics feature

Proof of concept of #361

### DIFF
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -365,13 +365,25 @@ export class LspServer {
         const geterrTokenSource = new lsp.CancellationTokenSource();
         this.diagnosticsTokenSource = geterrTokenSource;
 
-        const { files } = this.documents;
-        try {
-            return await this.tspClient.request(CommandTypes.Geterr, { delay: 0, files }, this.diagnosticsTokenSource.token);
-        } finally {
-            if (this.diagnosticsTokenSource === geterrTokenSource) {
-                this.diagnosticsTokenSource = undefined;
-                this.pendingDebouncedRequest = false;
+        if (this.workspaceConfiguration.diagnostics?.workspaceDiagnostics ?? false) {
+            const { files } = this.documents;
+            try {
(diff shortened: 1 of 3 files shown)


rchl:
Nice.

Likely this would need to run in a separate tsserver process as we can't allow a complication that could take minutes block all other functionality.

In vscode this feature is controlled with the `typescript.tsserver.experimental.enableProjectDiagnostics`, I believe. Though not sure how they've implemented it exactly. Would have to investigate it when I have some time.

rchl:
Quick check of typescript-language-features logs in VSCode shows that on every change it sends diagnostics for all files. Even files within `node_modules`. A bit inefficient if you ask me but might not be an issue in practice.

Still, it's dubbed as an "experimental" feature there so likely not fully optimized yet.

entropitor:
Yeah, if you have some pointers, I might have some time to implement this as it would benefit me greatly. Not sure how to best go from here 😄 I've already increased the delay to 1000 and then I didn't notice any performance problems myself. But I haven't used it a lot yet, if I clear the build cache, it might become unresponsive in the beginning. Potentially some debouncing (does the delay do this?) + separate process could help

Do you suggest using a seperate tsClient? Because I think it might use some of the configuration, ... so that might not fully work

rchl:
Without looking too deep into it, I think it should be implement with an additional tsClient. It would still receive the same initial configuration.

dimas-cyriaco:
Any chance this can be merged. This feature would be greatly appreciated.

rchl:
#585 will get us a step closer to this feature but there will still be a bunch more code that will have to be ported and adapted from VSCode.

I'll close this PR as it's the right way to do it.

weilbith:
> I'll close this PR as it's not the right way to do it.

@rchl as your linked PR has been merged, is there any follow up issue or PR for this I could follow?

rchl:
There is a bunch of work to be done to make that work and I can't justify working on that at the moment.

Some in-progress, non-working changes are on the https://github.com/typescript-language-server/typescript-language-server/tree/feat/project-diagnostics branch.

PR #371, opened 2022-01-13
fix: use correct name for the addMissingImports code action

rchl:


predragnikolic (approved):

src/utils/types.ts
```diff
@@ -12,7 +12,7 @@ export class CodeActionKind {
     public static readonly QuickFix = new CodeActionKind(lsp.CodeActionKind.QuickFix);
     public static readonly Refactor = new CodeActionKind(lsp.CodeActionKind.Refactor);
     public static readonly Source = new CodeActionKind(lsp.CodeActionKind.Source);
-    public static readonly SourceAddMissingImportsTs = CodeActionKind.Source.append('SourceAddMissingImportsTs').append('ts');
+    public static readonly SourceAddMissingImportsTs = CodeActionKind.Source.append('addMissingImports').append('ts');
```
predragnikolic, line 15: This message is just an observations. (not asking for any changes)

I haven't tested this, 
I am not sure how it worked with `SourceAddMissingImportsTs`...
or if it works with `addMissingImports`...

Can you put any string instead of `SourceAddMissingImportsTs`, and the add missing import will continue to work?

I looked for a code action on the typescript repo, and the closes thing that I could find is this name `fixMissingImport` :)
https://github.com/microsoft/TypeScript/blob/main/src/services/codefixes/importFixes.ts#L4

rchl: I'm not sure exactly what are you asking but this is matched against what the client is sending (in case of LSP the `lsp_code_actions_on_save` setting). If it matches then it will send code action if available from https://github.com/typescript-language-server/typescript-language-server/blob/f81303adfb7a795bcf215974c67f275f1af329f5/src/features/fix-all.ts#L164-L182

predragnikolic: That is the answer I was looking for :)

PR #372, opened 2022-01-16
fix: wait for tsserver configuration requests to finish

icholy:
Fix delayed `$/progress` begin event (https://github.com/typescript-language-server/typescript-language-server/pull/326). This is a work-around for https://github.com/microsoft/TypeScript/issues/47465. AFAIK, this is how vscode does it too.

Here's a demo of it correctly sending the "Initializing JS/TS language features ..." event.

tty

Here's what I used for testing https://gist.github.com/icholy/23670692d6405e676611fa49081a5f06


rchl:
Makes sense, thanks!

PR #373, opened 2022-01-16
feat: send $/progress reports every 500 milliseconds

icholy:
tsserver doesn't send progress reports when loading a project. This change sends a "In Progress" report every 500 milliseconds between `startedLoadingProject` and `finishedLoadingProject`. These are often used to update the UI in LSP clients. For example, [lsp_progress](https://github.com/arkav/lualine-lsp-progress) uses it to advance the spinner.

tty

I'm not convinced it's worth it, but I was curious what you thought.

### DIFF
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -43,6 +43,7 @@ import { CodeActionKind } from './utils/types';
 class ServerInitializingIndicator {
     private _loadingProjectName?: string;
     private _progressReporter?: ProgressReporter;
+    private _reportInterval?: NodeJS.Timeout;
 
     constructor(private lspClient: LspClient) {}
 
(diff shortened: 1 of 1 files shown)


rchl:
The "in progress" is implied by there being a progress notification started and lasts until it's ended.

The client are free to also show an "indefinite" progress bar for those since `percentage` is not provided.

So showing an additional "in progress" text seems redundant an unnecessary to me.

rchl:
For that specific editor case (lsp_progress), I think it would likely make sense for it to animate the progress bar on `percentage`-less progresses automatically. The server should not need to do something redundant to cater for such needs of a specific editor.

icholy:
Ya, I agree.

PR #377, opened 2022-01-17
fix: call configure before completion resolve 

fikuuss:
Auto import doesn't use insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces formatting option


rchl:
Can you add a test for it or at least provide a reproduction?

fikuuss:
Of course, no problem. It's already night for me now, so I'll add the tests tomorrow.

rchl (approved):

PR #547, opened 2022-07-25
refactor: ship as an ES module

rchl:
BREAKING CHANGE: Ship as an ES module. Might be breaking for programmatic users of this server. Read https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c


src/features/fix-all.ts
```diff
@@ -3,14 +3,15 @@
...
 
-import tsp, { CommandTypes } from 'typescript/lib/protocol';
-import * as lsp from 'vscode-languageserver/node';
-import { LspDocuments } from '../document';
-import { toFileRangeRequestArgs, toTextDocumentEdit } from '../protocol-translation';
-import { TspClient } from '../tsp-client';
-import * as errorCodes from '../utils/errorCodes';
-import * as fixNames from '../utils/fixNames';
-import { CodeActionKind } from '../utils/types';
+import type tsp from 'typescript/lib/protocol.d.js';
```
CGNonofr, line 6: Why do you need to import the d.ts? it seems uncommon

rchl: It would work if I'd use the `.js` extension but in this case the `protocol.js` file doesn't exist physically, only `protocol.d.ts` so this seems a bit more explicit in describing what we are doing.

CGNonofr: LGTM then

CGNonofr (approved):

CGNonofr:
Can I do something to help make this go forward to have the next release published?

PR #566, opened 2022-08-20
feat: support `textDocument/inlayHint` request from 3.17.0 spec

rchl:


src/features/inlay-hints.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        const document = documents.get(file);
+
+        if (!document) {
+            lspClient.showErrorMessage('Inlay Hints request failed. File not opened in the editor.');
+            return [];
+        }
+
+        if (!areInlayHintsEnabledForFile(configurationManager, file)) {
```
predragnikolic, line 53: is this check necessary?

src/features/inlay-hints.ts
```diff
@@ -0,0 +1,98 @@
...
+        const start = document.offsetAt(range.start);
+        const length = document.offsetAt(range.end) - start;
+
+        const response = await tspClient.request(CommandTypes.ProvideInlayHints, { file, start, length });
+        if (response.type !== 'response' || !response.success || !response.body) {
+            return [];
+        }
+
+        return response.body.map<lsp.InlayHint>(hint => ({
+            kind:  fromProtocolInlayHintKind(hint.kind),
```
predragnikolic, line 68: Want to follow the spec more precisely?

Currently the `fromProtocolInlayHintKind` can return a kind or undefined.
only add the kind to the inlay hints if it was returned from `fromProtocolInlayHintKind`

```
/**
	 * The kind of this hint. Can be omitted in which case the client
	 * should fall back to a reasonable default.
	 */
	kind?: [InlayHintKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#inlayHintKind);
```

predragnikolic:
Initial testing show that this works :heavy_check_mark: :+1: 

Do you see a need to add a check to respect client capabilities if the client supports inlayHints? (IMO I do not see a need this check, because the client will not send a inlayHints request in the first place if it doesn't support it)

src/features/inlay-hints.ts
rchl: Yes. Otherwise, even with all inlay hint options disabled, the server would still do expensive processing that takes a significant amount of time. The old implementation suffered from that which I've realized when working on this.

src/features/inlay-hints.ts
rchl: I suppose you are right. I've assumed that missing property is mostly equivalent to `undefined` value but in stricter implementations it might not be so.

This is based on VSCode's code which doesn't use LSP so it might be getting away with looser logic.

rchl:
> Do you see a need to add a check to respect client capabilities if the client supports inlayHints? (IMO I do not see a need this check, because the client will not send a inlayHints request in the first place if it doesn't support it)

No, I don't see the point. Would only be needed if the server would be the initiating side IMO.

PR #569, opened 2022-08-21
Move deepmerge from devDependency to runtime dependency

adamvy:
configuration-manager.ts requires deepmerge, but it's only listed as a devDependency.

PR #570, opened 2022-08-22
Move deepmerge to dependencies

polyzen:
The server fails to start without it:
`Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'deepmerge' imported from /usr/lib/node_modules/typescript-language-server/lib/configuration-manager.js`


polyzen:
Duplicate of #569

PR #579, opened 2022-09-04
feat: include "triggerReason" and "kind" in code action requests

rchl:


src/lsp-server.ts
```diff
@@ -906,14 +905,24 @@ export class LspServer {
...
         } catch (err) {
             return undefined;
         }
     }
-    protected async getRefactors(args: tsp.GetApplicableRefactorsRequestArgs): Promise<tsp.GetApplicableRefactorsResponse | undefined> {
+    protected async getRefactors(fileRangeArgs: tsp.FileRangeRequestArgs, context: lsp.CodeActionContext): Promise<tsp.GetApplicableRefactorsResponse | undefined> {
+        const args: tsp.GetApplicableRefactorsRequestArgs = {
+            ...fileRangeArgs,
+            triggerReason: context.triggerKind === lsp.CodeActionTriggerKind.Invoked ? 'invoked' : undefined,
+            kind: context.only?.length === 1 ? context.only[0] : undefined,
```
rchl, line 924: Typescript only have support for passing 1 `kind` while in the LSP world it's a list of `kinds` so I'm only respecting client's kind when there is only one. Alternatively we could filter kinds on our side I guess but for now I'll go with this safer (more backward-compatible) approach.

PR #583, opened 2022-09-05
feat: Add insert replace support for completions

predragnikolic:
This PR adds insert replace support.

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -63,7 +63,7 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
     }
 
     let insertText = entry.insertText;
-    const replacementSpan = entry.replacementSpan ?? optionalReplacementSpan;
+    const replacementSpan = entry.replacementSpan || optionalReplacementSpan;
     let replacementRange = replacementSpan && Range.fromTextSpan(replacementSpan);
     // Make sure we only replace a single line at most
     if (replacementRange && replacementRange.start.line !== replacementRange.end.line) {
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -116,7 +116,7 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
 }
 
 function createTextEdit(replacementRange: lsp.Range, position: lsp.Position, insertText: string, features: SupportedFeatures): lsp.TextEdit | lsp.InsertReplaceEdit {
-    if (features.insertReplaceSupport) {
+    if (features.completionInsertReplaceSupport) {
         const insertRange = lsp.Range.create(replacementRange.start, position);
         return lsp.InsertReplaceEdit.create(insertText, insertRange, replacementRange);
     }
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -105,13 +105,14 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
             }
         }
     }
-    if (!insertText) {
-        insertText = item.label;
-    }
     if (replacementRange) {
+        if (!insertText) {
+            insertText = item.label;
+        }
         item.textEdit = createTextEdit(replacementRange, position, insertText, features);
+    } else {
+        item.insertText = insertText;
     }
-    item.insertText = insertText;
     return item;
 }
 
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -109,19 +109,16 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
         if (!insertText) {
             insertText = item.label;
         }
-        item.textEdit = createTextEdit(replacementRange, position, insertText, features);
+        if (features.completionInsertReplaceSupport) {
+            const insertRange = lsp.Range.create(replacementRange.start, position);
+            item.textEdit = lsp.InsertReplaceEdit.create(insertText, insertRange, replacementRange);
+        } else {
+            item.textEdit = lsp.TextEdit.replace(replacementRange, insertText);
+        }
     } else {
         item.insertText = insertText;
     }
     return item;
-}
-
-function createTextEdit(replacementRange: lsp.Range, position: lsp.Position, insertText: string, features: SupportedFeatures): lsp.TextEdit | lsp.InsertReplaceEdit {
-    if (features.completionInsertReplaceSupport) {
-        const insertRange = lsp.Range.create(replacementRange.start, position);
-        return lsp.InsertReplaceEdit.create(insertText, insertRange, replacementRange);
-    }
-    return lsp.TextEdit.replace(replacementRange, insertText);
 }
 
 function asCompletionItemKind(kind: ScriptElementKind): lsp.CompletionItemKind {
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -22,7 +22,7 @@ interface ParameterListParts {
     readonly hasOptionalParameters: boolean;
 }
 
-export function asCompletionItem(entry: tsp.CompletionEntry, file: string, position: lsp.Position, document: LspDocument, features: SupportedFeatures, completionInfo?: tsp.CompletionInfo): lsp.CompletionItem | null {
+export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacementSpan: tsp.TextSpan | undefined, file: string, position: lsp.Position, document: LspDocument, features: SupportedFeatures): lsp.CompletionItem | null {
     const item: lsp.CompletionItem = {
         label: entry.name,
         ...features.completionLabelDetails ? { labelDetails: entry.labelDetails } : {},
@@ -63,7 +63,7 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
     }
 
     let insertText = entry.insertText;
-    const replacementSpan = entry.replacementSpan ?? completionInfo?.optionalReplacementSpan;
+    const replacementSpan = entry.replacementSpan ?? optionalReplacementSpan;
     let replacementRange = replacementSpan && Range.fromTextSpan(replacementSpan);
     // Make sure we only replace a single line at most
     if (replacementRange && replacementRange.start.line !== replacementRange.end.line) {
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -22,7 +22,7 @@ interface ParameterListParts {
     readonly hasOptionalParameters: boolean;
 }
 
-export function asCompletionItem(entry: tsp.CompletionEntry, file: string, position: lsp.Position, document: LspDocument, features: SupportedFeatures): lsp.CompletionItem | null {
+export function asCompletionItem(entry: tsp.CompletionEntry, file: string, position: lsp.Position, document: LspDocument, features: SupportedFeatures, completionInfo?: tsp.CompletionInfo): lsp.CompletionItem | null {
     const item: lsp.CompletionItem = {
         label: entry.name,
         ...features.completionLabelDetails ? { labelDetails: entry.labelDetails } : {},
@@ -63,7 +63,8 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
     }
 
     let insertText = entry.insertText;
-    let replacementRange = entry.replacementSpan && Range.fromTextSpan(entry.replacementSpan);
+    const replacementSpan = entry.replacementSpan ?? completionInfo?.optionalReplacementSpan;
+    let replacementRange = replacementSpan && Range.fromTextSpan(replacementSpan);
     // Make sure we only replace a single line at most
     if (replacementRange && replacementRange.start.line !== replacementRange.end.line) {
         replacementRange = lsp.Range.create(replacementRange.start, document.getLineEnd(replacementRange.start.line));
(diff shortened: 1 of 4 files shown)


src/completion.ts
```diff
@@ -104,18 +105,24 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
...
-        if (!insertText) {
-            insertText = item.label;
-        }
-
-        item.textEdit = lsp.TextEdit.replace(replacementRange, insertText);
-    } else {
-        item.insertText = insertText;
+        item.textEdit = createTextEdit(replacementRange, position, insertText, features);
     }
+    item.insertText = insertText;
```
predragnikolic, line 114: Should I put `item.insertText`  back in an else branch?

predragnikolic:
I had some issues with committing locally, 
```
husky > pre-commit (node v16.16.0)
$ eslint --ext ".js,.ts" src
$ concurrently -n compile,lint -c blue,green "yarn compile" "yarn lint"
[lint] $ eslint --ext ".js,.ts" src
[compile] $ tsc -b
[compile] yarn compile exited with code 0
[lint] yarn lint exited with code 0
$ cross-env CONSOLE_LOG_LEVEL=warn mocha "./lib/**/*.spec.js"
(node:97369) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:97369) ExperimentalWarning: The Node.js specifier resolution flag is experimental. It could change or be removed at any time.

ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/home/predragnikolic/Documents/sandbox/typescript-language-server/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///home/predragnikolic/Documents/sandbox/typescript-language-server/lib/modules-resolver.spec.js:27:23
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at importModuleDynamicallyWrapper (node:internal/vm/module:437:15)
    at formattedImport (/home/predragnikolic/Documents/sandbox/typescript-language-server/node_modules/mocha/lib/nodejs/esm-utils.js:7:14)
    at Object.exports.requireOrImport (/home/predragnikolic/Documents/sandbox/typescript-language-server/node_modules/mocha/lib/nodejs/esm-utils.js:38:28)
    at Object.exports.loadFilesAsync (/home/predragnikolic/Documents/sandbox/typescript-language-server/node_modules/mocha/lib/nodejs/esm-utils.js:91:20)
    at singleRun (/home/predragnikolic/Documents/sandbox/typescript-language-server/node_modules/mocha/lib/cli/run-helpers.js:125:3)
    at Object.exports.handler (/home/predragnikolic/Documents/sandbox/typescript-language-server/node_modules/mocha/lib/cli/run.js:370:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
husky > pre-commit hook failed (add --no-verify to bypass)
```
Probably because I am using Node v14.18.3, and in the logs I can see node v16.16.0
so I just committed with --no-verify. 

Is that Ok? 
Or should I take some actions and if so what exactly?

rchl:
Will repeat myself for the record:

I'm a bit concerned about doing that as it will add a ton of extra bytes to the response which can already be in megabytes...

On the other hand, if it's just the correct thing to do, we might have to...

rchl:
> so I just committed with --no-verify.
> 
> Is that Ok?

It's OK but you still have to fix the CI errors so you might as well just switch to Node 16. :)
Repo has `.nvmrc` so it will switch automatically if you are using nvm.

src/ts-protocol.ts
```diff
@@ -30,6 +30,7 @@ export interface SupportedFeatures {
     completionSnippets?: boolean;
     diagnosticsTagSupport?: boolean;
     definitionLinkSupport?: boolean;
+    insertReplaceSupport?: boolean;
```
rchl, line 33: `completionInsertReplaceSupport` to make it clear which feature it applies to.

src/lsp-server.ts
```diff
@@ -655,14 +656,15 @@ export class LspServer {
...
+                includeInsertTextCompletions: true,
             }));
             const { body } = result;
             const completions: lsp.CompletionItem[] = [];
             for (const entry of body?.entries ?? []) {
                 if (entry.kind === 'warning') {
                     continue;
                 }
-                const completion = asCompletionItem(entry, file, params.position, document, this.features);
+                const completion = asCompletionItem(entry, file, params.position, document, this.features, body);
```
rchl, line 667: It's weird to pass just a single `entry` as one argument and then also the whole response as another. I'd prefer that you just pass `body.optionalReplacementSpan`. And you can move the argument after `entry` maybe since it kind of belongs better there.

src/lsp-server.ts
```diff
@@ -655,14 +656,15 @@ export class LspServer {
                 offset: params.position.character + 1,
                 triggerCharacter: getCompletionTriggerCharacter(params.context?.triggerCharacter),
                 triggerKind: params.context?.triggerKind,
+                includeInsertTextCompletions: true,
```
rchl, line 659: Do you know if this is still relevant if client does not support InsertReplace? Maybe we should make it conditional on the capability if that will make us save some cycles and bytes?

src/lsp-server.ts
```diff
@@ -188,9 +188,10 @@ export class LspServer {
         // Setup supported features.
         const { textDocument } = clientCapabilities;
         if (textDocument) {
+            const completionCapabilities = textDocument.completion;
             this.features.codeActionDisabledSupport = textDocument.codeAction?.disabledSupport;
             this.features.definitionLinkSupport = textDocument.definition?.linkSupport && typescriptVersion.version?.gte(API.v270);
-            const completionCapabilities = textDocument.completion;
+            this.features.insertReplaceSupport = completionCapabilities?.completionItem?.insertReplaceSupport ?? false;
```
rchl, line 194: Fine if the value is `undefined`:

```suggestion
            this.features.insertReplaceSupport = completionCapabilities?.completionItem?.insertReplaceSupport;
```

src/completion.ts
```diff
@@ -63,7 +63,8 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
     }
 
     let insertText = entry.insertText;
-    let replacementRange = entry.replacementSpan && Range.fromTextSpan(entry.replacementSpan);
+    const replacementSpan = entry.replacementSpan ?? completionInfo?.optionalReplacementSpan;
```
rchl, line 66: Fine with using `||` here.

src/completion.ts
rchl: We should keep the logic as close to how it was before as possible.
Maybe this change is the reasons the tests are failing even?

src/completion.ts
```diff
@@ -104,18 +105,24 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
...
-        item.insertText = insertText;
+        item.textEdit = createTextEdit(replacementRange, position, insertText, features);
     }
+    item.insertText = insertText;
     return item;
 }
 
+function createTextEdit(replacementRange: lsp.Range, position: lsp.Position, insertText: string, features: SupportedFeatures): lsp.TextEdit | lsp.InsertReplaceEdit {
+    if (features.insertReplaceSupport) {
+        const insertRange: lsp.Range = { start: replacementRange.start, end: position };
```
rchl, line 120: I'm not sure there is really a need for a separate function for this but it would be simpler to use helper function here.

```suggestion
        const insertRange = lsp.Range.create(replacementRange.start, position);
```

predragnikolic:
> I'm a bit concerned about doing that as it will add a ton of extra bytes to the response which can already be in megabytes...

LSP spec to the rescue :)
They added something new to the CompletionList type.  

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList

```
/**
	 * In many cases the items of an actual completion result share the same
	 * value for properties like `commitCharacters` or the range of a text
	 * edit. A completion list can therefore define item defaults which will
	 * be used if a completion item itself doesn't specify the value.
	 *
	 * If a completion list specifies a default value and a completion item
	 * also specifies a corresponding value the one from the item is used.
	 *
	 * Servers are only allowed to return default values if the client
	 * signals support for this via the `completionList.itemDefaults`
	 * capability.
	 *
	 * @since 3.17.0
	 */
	itemDefaults?: {
		/**
		 * A default commit character set.
		 *
		 * @since 3.17.0
		 */
		commitCharacters?: string[];

		/**
		 * A default edit range
		 *
		 * @since 3.17.0
		 */
		editRange?: [Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range) | {
			insert: [Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range);
			replace: [Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range);
		};

		/**
		 * A default insert text format
		 *
		 * @since 3.17.0
		 */
		insertTextFormat?: [InsertTextFormat](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertTextFormat);

		/**
		 * A default insert text mode
		 *
		 * @since 3.17.0
		 */
		insertTextMode?: [InsertTextMode](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#insertTextMode);

		/**
		 * A default data value.
		 *
		 * @since 3.17.0
		 */
		data?: [LSPAny](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#lspAny);
	}
```

src/lsp-server.ts
predragnikolic: this is probably not necessary and `includeInsertTextCompletions` is deprecated, 
I tried everything just to make this work :D and forgot to remove this.

predragnikolic:
Re-guarding the new spec `itemDefaults` I do not think that I will implement it as part of this PR.
I'll continue with the PR. 
The itemDefaults is a really nice addition to the spec, but I suppose there will be a future PR that will add support for that.

src/lsp-server.spec.ts
```diff
@@ -201,6 +201,70 @@ describe('completion', () => {
...
+        const clientCapabilitiesOverride: lsp.ClientCapabilities = {
+            textDocument: {
+                completion: {
+                    completionItem: {
+                        insertReplaceSupport: true,
+                    },
+                },
+            },
+        };
+        const localServer = await createServer({
```
predragnikolic, line 238: One thing to note,
not sure if it is the best way to create a localServer here and to clean up at the end of this function.
This test is taking ~5s .

src/lsp-server.spec.ts
```diff
@@ -201,6 +201,70 @@ describe('completion', () => {
...
+        localServer.didOpenTextDocument({ textDocument: doc });
+        const proposals = await localServer.completion({ textDocument: doc, position: positionAfter(doc, '.get') });
+        assert.isNotNull(proposals);
+        const completion = proposals!.items.find(completion => completion.label === 'getById');
+        assert.isDefined(completion);
+        assert.isDefined(completion!.textEdit);
+        assert.containsAllKeys(completion!.textEdit, ['newText', 'insert', 'replace']);
+        localServer.didCloseTextDocument({ textDocument: doc });
+        localServer.closeAll();
+        localServer.shutdown();
```
predragnikolic, line 265: Did I miss to clean up something? (diagnostics?)

rchl: You can just provide empty `publishDiagnostics` implementation if you don't need to check diagnostics.

predragnikolic: I do not need to check diagnostics.
Still unsure exactly I should put. 
Could you push that peace of code?

src/lsp-server.spec.ts
rchl: It's what it is :)

src/completion.ts
```diff
@@ -108,14 +109,21 @@ export function asCompletionItem(entry: tsp.CompletionEntry, file: string, posit
...
     return item;
 }
 
+function createTextEdit(replacementRange: lsp.Range, position: lsp.Position, insertText: string, features: SupportedFeatures): lsp.TextEdit | lsp.InsertReplaceEdit {
+    if (features.completionInsertReplaceSupport) {
+        const insertRange = lsp.Range.create(replacementRange.start, position);
+        return lsp.InsertReplaceEdit.create(insertText, insertRange, replacementRange);
+    }
+    return lsp.TextEdit.replace(replacementRange, insertText);
+}
```
rchl, line 125: As I've alluded earlier, I don't think this function adds value and it just adds extra unnecessary lines.
We can inline this code.

rchl (approved):

rchl:
Thanks!

rchl:
Been running with it for a while and I'm a bit annoyed by the new behavior (this is without corresponding LSP changes).

For a code like:

```js
return path.reso|scriptArgs[0] || './.lokalise';
```

where `|` is the cursor it will replace the following text and end up with 


```js
return path.resolve[0] || './.lokalise';
```

I'm not sure if I'm not used to it or it's just annoying...

rchl:
I don't know, that would be also a behavior change and potentially breaking. It's just that previously this happened only in specific cases while now it happens (almost?) every time.

Alternatively we could only respect `optionalReplacementSpan` when client supports InsertReplace.

predragnikolic:
> Alternatively we could only respect/use optionalReplacementSpan when client supports InsertReplace.

Yes, that is the best way to keep the prev behavior.

PR #584, opened 2022-09-10
fix: only use optionalReplacementSpan if client supports InsertReplace

rchl:
@predragnikolic

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -63,7 +63,7 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
     }
 
     let insertText = entry.insertText;
-    const replacementSpan = entry.replacementSpan || optionalReplacementSpan;
+    const replacementSpan = entry.replacementSpan || (features.completionInsertReplaceSupport ? optionalReplacementSpan : undefined);
     let replacementRange = replacementSpan && Range.fromTextSpan(replacementSpan);
     // Make sure we only replace a single line at most
     if (replacementRange && replacementRange.start.line !== replacementRange.end.line) {
--- a/src/lsp-server.spec.ts
+++ b/src/lsp-server.spec.ts
@@ -201,7 +201,7 @@ describe('completion', () => {
         server.didCloseTextDocument({ textDocument: doc });
     });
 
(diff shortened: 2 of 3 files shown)


src/completion.ts
```diff
@@ -63,7 +63,7 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
     }
 
     let insertText = entry.insertText;
-    const replacementSpan = entry.replacementSpan || optionalReplacementSpan;
+    const replacementSpan = entry.replacementSpan || (features.completionInsertReplaceSupport ? optionalReplacementSpan : undefined);
```
predragnikolic, line 66: I would suggest removing unnecessary code bellow:

```diff
- const replacementSpan = entry.replacementSpan || (features.completionInsertReplaceSupport ? optionalReplacementSpan : undefined);
+ const replacementSpan = entry.replacementSpan || optionalReplacementSpan;

...

- if (replacementRange) {
+ if (replacementRange && features.completionInsertReplaceSupport) {
    if (!insertText) {
        insertText = item.label;
    }
-    if (features.completionInsertReplaceSupport) {
      const insertRange = lsp.Range.create(replacementRange.start, position);
      item.textEdit = lsp.InsertReplaceEdit.create(insertText, insertRange, replacementRange);
-    } else {
-        item.textEdit = lsp.TextEdit.replace(replacementRange, insertText);
-    }
} else {
    item.insertText = insertText;
}
```

rchl: That doesn't sound right. The replacementRange can also come from the `entry.replaceSpan` in which case we want to create a replace TextEdit.

predragnikolic: True, you are right.

predragnikolic: Will try out this PR tomorrow.

predragnikolic:
Just to document that this PR addresses this comment https://github.com/typescript-language-server/typescript-language-server/pull/583#issuecomment-1241795204

predragnikolic (approved):
I can verify that this PR brings back the same behavior as prior to InsertReplaceSupport for servers that do not have `insertReplaceSupport`.

Issue #358, opened 2022-01-04
Show all errors and warnings in project for all Typescript files, not just opened ones


niebloomj:
We do see all errors and warnings in the "problems" tab when we open a Typescript file with errors, but they are removed from the problems tab when we close the file. It would be great to be able to see all the problems of all the files in a project.

This was requested to be globally implemented in the vscode issue at microsoft/vscode#13953 but we were told that this is not something to be implemented globally, but per language/plugin.

I do understand not everyone would like to have this feature, or that it would be useful just for some cases or projects., specially because it would consume too many resources. A way to turn this feature on/off per project would help the user when to turn it on/off.

rchl:
You seem to be using VSCode and vscode doesn't use this language server. It has its own implementation that doesn't use the LSP protocol at https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features

That said, it was discussed previously in #253 in context of non-VSCode editors. The conclusion was that something like this is better implemented in the client (so opposite of the discussion you've linked ;)).

Even if we would decide that it belongs in the language server, I don't think the `tsserver` provides an API for compiling the whole project. So before anything could be done here, the typescript project would have to add that first. Though I imagine that would have to run in a separate process so that it doesn't block all the existing functionality as it would likely require quite a bit of time to compile some projects.

niebloomj:
Thank you. I'll keep digging into this so I can figure out how we can best solve this issue. Thank you for your comment!

entropitor:
@rchl And what if this is an option so clients have to opt-in (or even send a notification which will trigger sending all workspace diagnostics)? 

The first build can take a while but after that, if you have project references + incremental builds turned on, builds can be super fast so I think it can be feasible for those cases and it would be a major win as seeing all errors in your editor is so useful!

rchl:
There could be an option (typescript project usually adds options for the feature it implements anyway). But in any case, it should be discussed in the [typescript](https://github.com/microsoft/TypeScript) repo first IMO to figure out the best way forward.

entropitor:
It seems to already be in tsserver: https://github.com/typescript-language-server/typescript-language-server/pull/361 This seems to be quite slow though so we still need to do it in the background and make sure we run it for all projects that are open but it seems to work already

Rigo-m:
Any news on this? Unfortunately I'm new to this stuff, just started using nvim from vscode and cannot be of great help.
If someone has a working workaround i'd be glad <3

[...]

PR #585, opened 2022-09-11
feat: support communicating with tsserver using IPC

rchl:
Ported a bunch of code from "VSCode typescript features" to enable us to be able to communicate with `tsserver` over IPC instead of `stdio`. IPC is used for tsserver (typescript) versions 4.60+, otherwise the legacy stdio is used.

If this seems like a lot of code to handle that case then yes, it is, but it's part of the plan to align the code as closely as possible with VSCode to be able to support features like #358 without re-inventing the wheel.

### DIFF
--- a/src/tsServer/cancellation.ts
+++ b/src/tsServer/cancellation.ts
@@ -1,0 +1,67 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+/*
+ * Copyright (C) 2022 TypeFox and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+import fs from 'node:fs';
+import { temporaryFile } from 'tempy';
+// import Tracer from '../utils/tracer';
+
+export interface OngoingRequestCanceller {
+    readonly cancellationPipeName: string | undefined;
+    tryCancelOngoingRequest(seq: number): boolean;
+}
+
+export interface OngoingRequestCancellerFactory {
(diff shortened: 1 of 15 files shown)


niebloomj:
I love the idea here! Anything others can do to help?

src/tsServer/cancellation.ts
```diff
@@ -0,0 +1,67 @@
...
+            fs.writeFileSync(this.cancellationPipeName + String(seq), '');
+        } catch {
+            // noop
+        }
+        return true;
+    }
+}
+
+export const nodeRequestCancellerFactory = new class implements OngoingRequestCancellerFactory {
+    create(/*serverId: string, tracer: Tracer*/): OngoingRequestCanceller {
```
predragnikolic, line 64: should `/*serverId: string */` be commented out?

rchl: It should as it's not used currently. But forgot to comment out in the interfaces also.

predragnikolic: It looks like serverId is not necessary or used in `NodeRequestCanceller,`
but it just looked strange to see `serverId` being commented out only for this method.

src/tsServer/spawner.ts
```diff
@@ -0,0 +1,154 @@
...
+import API from '../utils/api.js';
+import { ServerType } from './requests.js';
+import { Logger } from '../logger.js';
+import type { TspClientOptions } from '../tsp-client.js';
+import { nodeRequestCancellerFactory } from './cancellation.js';
+import { ITypeScriptServer, ProcessBasedTsServer, TsServerProcessKind } from './server.js';
+import { NodeTsServerProcessFactory } from './serverProcess.js';
+import type { TypeScriptVersion } from './versionProvider.js';
+
+export class TypeScriptServerSpawner {
```
predragnikolic, line 21: Is it maybe planned to have multiple processes for each `TsServerProcessKind`?

```
export const enum TsServerProcessKind {
    Main = 'main',
    Syntax = 'syntax',
    Semantic = 'semantic',
    Diagnostics = 'diagnostics'
}
```

rchl: Not sure if multiple for each but I don't think so...

It would be one of each at most, I imagine.

PR #592, opened 2022-09-25
fix: wrong import completion when insert/replace supported

rchl:
Fixes regression for clients that have `insertReplaceSupport` where completing:

```ts
import { readFil }
```

would result in incorrect insert:

```ts
import { readFile } from "fs"; }
```

The TextEdit in this case would be:

```
      "textEdit": {
        "insert": {
          "end": {
            "character": 16,
            "line": 0
          },
          "start": {
            "character": 0,
            "line": 0
          }
        },
        "newText": "import { readFile$1 } from \"fs\";",
        "replace": {
          "end": {
            "character": 18,
            "line": 0
          },
          "start": {
            "character": 0,
            "line": 0
          }
        }
      }
```

and the problem would be in the `insert` range.

We should not be providing both insert and replace ranges when TS completion entry has its own `replacementSpan` provided as in that case it's the only valid option.

@predragnikolic

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -50,11 +50,11 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
         item.sortText = '\uffff' + entry.sortText;
     }
 
-    const { isSnippet, sourceDisplay } = entry;
+    const { isSnippet, replacementSpan, sourceDisplay } = entry;
     if (isSnippet && !features.completionSnippets) {
         return null;
     }
-    if (features.completionSnippets && (isSnippet || entry.isImportStatementCompletion || item.kind === lsp.CompletionItemKind.Function || item.kind === lsp.CompletionItemKind.Method)) {
+    if (features.completionSnippets && (isSnippet || item.kind === lsp.CompletionItemKind.Function || item.kind === lsp.CompletionItemKind.Method)) {
         // Import statements, Functions and Methods can result in a snippet completion when resolved.
         item.insertTextFormat = lsp.InsertTextFormat.Snippet;
     }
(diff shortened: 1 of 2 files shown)


src/completion.ts
```diff
@@ -50,28 +50,30 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
         item.sortText = '\uffff' + entry.sortText;
     }
 
-    const { isSnippet, sourceDisplay } = entry;
+    const { isSnippet, replacementSpan, sourceDisplay } = entry;
     if (isSnippet && !features.completionSnippets) {
         return null;
     }
-    if (features.completionSnippets && (isSnippet || entry.isImportStatementCompletion || item.kind === lsp.CompletionItemKind.Function || item.kind === lsp.CompletionItemKind.Method)) {
+    if (features.completionSnippets && (isSnippet || item.kind === lsp.CompletionItemKind.Function || item.kind === lsp.CompletionItemKind.Method)) {
```
rchl, line 57: The `entry.isImportStatementCompletion` check was removed it as it's only provided for telemetry reasons and `isSnippet` check covers that case already.

src/completion.ts
```diff
@@ -105,22 +107,34 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
...
-        } else {
-            item.textEdit = lsp.TextEdit.replace(replacementRange, insertText);
-        }
-    } else {
-        item.insertText = insertText;
-    }
     return item;
 }
 
+function getRangeFromReplacementSpan(
```
predragnikolic, line 113: this function is pretty similar to what VS Code does:

```ts
    // vscode/extensions/typescript-language-features/src/languageFeatures/completions.ts
	private getRangeFromReplacementSpan(tsEntry: Proto.CompletionEntry, completionContext: CompletionContext) {
		if (!tsEntry.replacementSpan) {
			return;
		}

		let replaceRange = typeConverters.Range.fromTextSpan(tsEntry.replacementSpan);
		// Make sure we only replace a single line at most
		if (!replaceRange.isSingleLine) {
			replaceRange = new vscode.Range(replaceRange.start.line, replaceRange.start.character, replaceRange.start.line, completionContext.line.length);
		}

		// If TS returns an explicit replacement range, we should use it for both types of completion
		return {
			inserting: replaceRange,
			replacing: replaceRange,
		};
	}
```

So basically if an entry has a replacementSpan, VS Code uses that replacementSpan for both insert and replace.

You took a little different approach, but it works.

rchl: VSCode doesn't use LSP so we can't use exactly same code but yes, it's based on VSCode's logic.

PR #593, opened 2022-09-26
fix(completions): remove filterText override for bracket accessor

rchl:
Removed this old code that overridden filterText for bracket accessor completions.

This old code originally came from VSCode and was later replaced there with a quite different code: https://github.com/microsoft/vscode/commit/beb5998ea3d6b4cae43b3a09ecdf11e3bd692667 to fix issue https://github.com/microsoft/vscode/issues/63100

I've tested in ST and don't see the issue that VSCode had so given no evidence that it's needed outside of VSCode, I'm not trying to replicate what they did here.

Also, in ST at least, the results are better without this old code:

Before:

Screenshot 2022-09-26 at 13 01 25

After:

Screenshot 2022-09-26 at 13 01 06

So before this change editor has shown a leading dot which is not really needed or expected.

As a bonus (in ST at least) it makes this case better as we can see that the property is optional:

Screenshot 2022-09-26 at 13 13 14

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -63,17 +63,6 @@ export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacement
     }
 
     let { insertText } = entry;
-    if (insertText && replacementSpan && insertText[0] === '[') { // o.x -> o['x']
-        item.filterText = '.' + item.label;
-    }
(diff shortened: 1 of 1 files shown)


predragnikolic (approved):
👍

PR #595, opened 2022-09-26
fix(completions): don't create snippet kind without `completeFunctionCalls`

rchl:
Server was supposed to create snippet completions on "resolve" for `method` and `function` completion kinds when the `completeFunctionCalls` workspace settings was enabled.

The code that has created the snippet during "resolve" was checking before-mentioned option but the code that initially set the `insertTextFormat` to `Snippet` did not. That meant that with `completeFunctionCalls` disabled, the `method` and `function` completions were set to `Snippet` kind initially but without a chance of being changed to actual snippets during "resolve".

And for editors that can't use resolved values that meant that those would potentially receive completions with unescaped `$` character in them while not actually being actual snippet.

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -22,7 +22,7 @@ interface ParameterListParts {
     readonly hasOptionalParameters: boolean;
 }
 
-export function asCompletionItem(entry: tsp.CompletionEntry, optionalReplacementSpan: tsp.TextSpan | undefined, file: string, position: lsp.Position, document: LspDocument, features: SupportedFeatures): lsp.CompletionItem | null {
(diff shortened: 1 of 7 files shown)


src/completion.ts
```diff
@@ -254,6 +254,10 @@ export async function isValidFunctionCompletionContext(filepath: string, positio
     }
 }
 
+function canCreateSnippetOfFunctionCall(kind: lsp.CompletionItemKind | undefined, options: WorkspaceConfigurationCompletionOptions): boolean {
+    return options.completeFunctionCalls === true && (kind === lsp.CompletionItemKind.Function || kind === lsp.CompletionItemKind.Method);
```
predragnikolic, line 258:
```suggestion
    return options.completeFunctionCalls && (kind === lsp.CompletionItemKind.Function || kind === lsp.CompletionItemKind.Method);
```

rchl: That would cause type error as `options.completeFunctionCalls` can be undefined and the function returns boolean.

PR #600, opened 2022-10-07
fix: disable IPC communication until TypeScript bug is fixed

### LINKED ISSUE #599: Multiple Instances of `tsserver.js` and `typingsInstaller.js` will be created every time `:PackerCompile` is run.
As the title suggests, I am running `typescript-language-server` inside `Neovim` with `packer.nvim` as my package manager. I noticed that every time I run `:PackerCompile`, the `tsserver.js` and `typingsInstaller.js` duplicate themselves. Worst, these processes remained in memory even after I quit Neovim; I had to kill them directly either by using Htop or by sending a SIGTERM to it. 

**These are the exact steps I used to reproduce this behavior:** 
1) Open up htop, and then set filter (pressing F4) to see only processes starting with 'node'
2) Launch neovim and open up any .js file
3) Run `:PackerCompile` from inside Neovim. 
(At this point, we can easily observe via htop `typingInstaller.js` and `tsserver.js` duplicating itself)
4) Quit Neovim with `:q`
(Observe that the `typingInstaller.js` and `tsserver.js` processes are still running in the background)

**Additional Info:**
I first noticed this strange behavior when I was editing my plugin config file while a .js buffer is opened. And because I set up an autocommand to run `:PackerCompile` every time my config file is saved, I noticed (when I was done tweaking my config) that my RAM usage suddenly spiked to about 8GB, which rarely occurs even during heavy web-browsing. When I fired up htop, I was shocked to find up to nearly twenty instances of tsserver.js running in the background. And after some experimentation, I was able to figure out that `:PackerCompile` is the culprit. 

I put the config for all my Neovim plugins inside one file: `~/.config/nvim/lua/plugins.lua`:
```lua
--Plugin management
local packer = require('packer')
vim.cmd([[
    augroup packer_user_config
	autocmd!
	autocmd BufWritePost plugins.lua source <afile> | PackerCompile
    augroup end
]])

local plugins = packer.startup({function() 
    --irrelevant lines omitted
    use
    {
        "neovim/nvim-lspconfig",
        config = function()
              local lspconfig = require("lspconfig")
              -- more lines omitted
              lspconfig.pyright.setup {
                  --omitted lines
              }
              lspconfig.tsserver.setup {
               
              }
              --setup for other language servers
        end
    }
end,
config = {...}
})

return plugins
```
**Please note that only `typescript-language-server` has this issue. I ran `:PackerCompile` while other language servers are running (pyright, sumnekolua, etc), and none of them exhibit this behavior.**

rchl:
The issue with using IPC communication for communicating with `tsserver` is that if the parent process gets killed then the `tsserver` process does not shut itself down.

This will be re-enabled once the Typescript issue (https://github.com/microsoft/TypeScript/issues/51100) is fixed.

Fixes #599

### DIFF
--- a/src/tsServer/serverProcess.ts
+++ b/src/tsServer/serverProcess.ts
@@ -15,7 +15,7 @@ import type { Readable } from 'node:stream';
 import type tsp from 'typescript/lib/protocol.d.js';
 import { TsServerProcess, TsServerProcessFactory, TsServerProcessKind } from './server.js';
 import type { TspClientOptions } from '../tsp-client.js';
-import API from '../utils/api.js';
+// import API from '../utils/api.js';
 import type { TypeScriptVersion } from './versionProvider.js';
 
 export class NodeTsServerProcessFactory implements TsServerProcessFactory {
@@ -26,7 +26,7 @@ export class NodeTsServerProcessFactory implements TsServerProcessFactory {
         configuration: TspClientOptions,
     ): TsServerProcess {
         const tsServerPath = version.tsServerPath;
-        const useIpc = version.version?.gte(API.v460);
+        const useIpc = false;  // version.version?.gte(API.v460);
 
         const runtimeArgs = [...args];
         if (useIpc) {
--- a/src/tsServer/spawner.ts
+++ b/src/tsServer/spawner.ts
@@ -47,12 +47,12 @@ export class TypeScriptServerSpawner {
             }
         }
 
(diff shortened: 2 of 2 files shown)


Ristovski:
Seems to be fixed upstream? https://github.com/microsoft/TypeScript/pull/51107

rchl:
I know. I've submitted the fix there :)

Now we need to wait for new version of TypeScript that includes the fix.

PR #607, opened 2022-10-11
feat: add `_typescript.configurePlugin` workspace command

mskelton:
This adds support for the `_typescript.configurePlugin` command which allows clients to dynamically update plugin configuration.  This is useful when configuring global plugins which, since defined via the command line flags, cannot be configured the same way they might when defined in a local `tsconfig.json` file.

An example of this in Neovim which enables the `typescript-styled-plugin` with some configuration changes.

```lua
require("lspconfig").tsserver.setup({
  init_options = {
    plugins = {
      {
        name = "typescript-styled-plugin",
        location = "..."
      }
    }
  },
  on_init = function()
    local params = {
      command = "_typescript.configurePlugin",
      arguments = {
        "typescript-styled-plugin",
        {
          validate = false,
          emmet = {
            showExpandedAbbreviation = false
          }
        }
      },
    }
    
    client.request("workspace/executeCommand", params)
  end
})
```

### DIFF
--- a/src/configuration-manager.ts
+++ b/src/configuration-manager.ts
@@ -116,6 +116,14 @@ export class ConfigurationManager {
         await this.tspClient?.request(CommandTypes.Configure, args);
     }
 
+    public async configurePlugin(pluginName: string, configuration: unknown): Promise<void> {
+        const args: tsp.ConfigurePluginRequestArguments = {
+            configuration,
+            pluginName,
+        };
+        await this.tspClient?.request(CommandTypes.ConfigurePlugin, args);
+    }
+
     public getPreferences(filename: string): tsp.UserPreferences {
         if (this.tspClient?.apiVersion.lt(API.v290)) {
             return {};
--- a/src/configuration-manager.ts
+++ b/src/configuration-manager.ts
@@ -116,14 +116,6 @@ export class ConfigurationManager {
         await this.tspClient?.request(CommandTypes.Configure, args);
     }
 
-    public async configurePlugin(pluginName: string, configuration: unknown): Promise<void> {
-        const args: tsp.ConfigurePluginRequestArguments = {
-            configuration,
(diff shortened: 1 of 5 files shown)


src/configuration-manager.ts
```diff
@@ -116,6 +116,14 @@ export class ConfigurationManager {
         await this.tspClient?.request(CommandTypes.Configure, args);
     }
 
+    public async configurePlugin(pluginName: string, configuration: unknown): Promise<void> {
+        const args: tsp.ConfigurePluginRequestArguments = {
+            configuration,
+            pluginName,
+        };
+        await this.tspClient?.request(CommandTypes.ConfigurePlugin, args);
```
rchl, line 124: You could follow vscode a bit more closely:

 - use `executeWithoutWaitingForResponse`
 - add API version check
 - make the function sync

https://github.com/microsoft/vscode/blob/97c8a97261377a11dc5d6b547e275f4cb4e9399a/extensions/typescript-language-features/src/typescriptServiceClient.ts#L1007-L1012

rchl, line 124: I like compact code so would suggest slight style change:

```suggestion
        await this.tspClient?.request(CommandTypes.ConfigurePlugin, { pluginName, configuration });
```

rchl: And to be fair, since it becomes a one-liner and it only used from one place, I would just inline it where the command is handled.

rchl: And to be fair, since it becomes a one-liner and it's only used from one place, I would just inline it where the command is handled.

mskelton: Thanks for the feedback, I've updated per your suggestions.

rchl (approved):

rchl:
Thanks

PR #610, opened 2022-10-14
fix: respect user-provided tsserver.js path from `--tsserver-path`

rchl:
Previous logic disallowed one to pass full path to an existing tsserver.js like for example a locally built one at `../built/local/tsserver.js`.


Esgariot:
Why the `[deprecated]` remark? Is there a new way for language server to discover tsserver.js?

PR #622, opened 2022-11-14
Prevent exception when tsserver crashes

stephank:
I was debugging an issue where tsserver crashes, and noticed the langserver also crashes because of a null dereference here. The exact scenario is a bit difficult to reproduce, because I'm debugging an issue in the patch Yarn v3 applies to TypeScript, and involves a very specific setup where langserver is a global install using PnP, but TypeScript is project local install using node_modules.

It was also a bit difficult to surface the exception in tsserver. I tried modifying the subprocess `stdio` config to inherit stderr (which I think is something we should probably do any way?), but it didn't seem to help. (I'm using neovim, and was hoping to see the exception in  `:LspLog` perhaps, but it only shows the 'exit code 1' message.)

### DIFF
--- a/src/tsServer/serverProcess.ts
+++ b/src/tsServer/serverProcess.ts
@@ -147,7 +147,7 @@ class StdioChildServerProcess implements TsServerProcess {
 
     kill(): void {
         this._process.kill();
-        this.reader.dispose();
+        this.reader?.dispose();
         this._reader = null;
     }
 }


rchl:
Do you know of any project that I could reproduce this issue on?
I would like to see if a fix can be made closer to the root issue as it seems like `kill` is being called twice in this case.

rchl:
Also note that with the next version of Typescript we'll be able to switch this server to communicate using IPC so I would like to also test that.

rchl:
I think I was able to reproduce this crash by manually breaking tsserver code.

I've created https://github.com/typescript-language-server/typescript-language-server/pull/624 to surface tsserver's `stderr` output and will also add a fix there to not call `kill` twice when that happens.

rchl:
Also added a change there to explicitly crash the tsserver process on tsserver crashing rather than crashing on null reference. typescript-language-server is not currently designed to recover from tsserver crash so I think it makes sense.

PR #646, opened 2022-12-27
fix(completions): don't insert call snippet if already a call

jadestrong:
https://github.com/microsoft/vscode/issues/18131

Align with the implementation of vscode:

https://github.com/microsoft/vscode/blob/6215269681d1afe10ddd786567d9753847213229/extensions/typescript-language-features/src/languageFeatures/completions.ts#L280-L299

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -235,14 +235,13 @@ export async function asResolvedCompletionItem(
         item.additionalTextEdits = asAdditionalTextEdits(details.codeActions, filepath);
         item.command = asCommand(details.codeActions, item.data.file);
     }
-    if (document) {
-        if (features.completionSnippets && canCreateSnippetOfFunctionCall(item.kind, options)) {
-            const { line, offset } = item.data;
-            const position = Position.fromLocation({ line, offset });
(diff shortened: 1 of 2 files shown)


src/completion.ts
```diff
@@ -234,41 +235,43 @@ export async function asResolvedCompletionItem(
...
         item.command = asCommand(details.codeActions, item.data.file);
     }
-    if (features.completionSnippets && canCreateSnippetOfFunctionCall(item.kind, options)) {
-        const { line, offset } = item.data;
-        const position = Position.fromLocation({ line, offset });
-        const shouldCompleteFunction = await isValidFunctionCompletionContext(filepath, position, client);
-        if (shouldCompleteFunction) {
-            createSnippetOfFunctionCall(item, details);
+    if (document) {
+        if (features.completionSnippets && canCreateSnippetOfFunctionCall(item.kind, options)) {
```
rchl, line 239: Can you combine those conditions to avoid creating extra indentation level?

jadestrong: done

rchl (approved):

PR #647, opened 2022-12-27
feat!: drop experimental `textDocument/calls`, `typescript/inlayHints`

rchl:
BREAKING CHANGE: Remove experimental and legacy implementations of inlay
  hints and call hierarchy. Switch to the official `textDocument/inlayHint`
  and `textDocument/prepareCallHierarchy` implementations instead.


rchl:
I will add `textDocument/prepareCallHierarchy` implementation later so there will be a replacement for removed `textDocument/calls`.

PR #669, opened 2023-01-21
Restore tsc version logging

tapayne88:
Hi 👋 

Thanks for the great work!

I think as part of https://github.com/typescript-language-server/typescript-language-server/pull/590 the way the typescript version is being reported was changed from `logIgnoringVerbosity` to `log`.
https://github.com/typescript-language-server/typescript-language-server/blob/d8b4cb698b6b74f12edb0a19e6c28b391cc0e985/src/lsp-server.ts#L106

Running typescript-language-server at the default log level now means we don't get these messages. My guess is this was done to stop the test producing `console.error`s so I've tried to restore the logging functionality while maintaining that error suppression.


rchl:
You are correct that I've disabled it due to `jest` being shouty. But another argument I had was that stderr is just generally not the right place to report stuff and is very specific to the communication method that the server uses which is stdio. If it would allow using IPC then those wouldn't show up at all.

That said, I am willing to accept this change but I would like you to consider an alternative of using `window/logMessage` notification (`this.options.lspClient.logMessage`). Would that work for your specific editor? I guess the answer can vary per editor, depending how those handle stderr messages vs. standard LSP log messages...

tapayne88:
Ah interest, that looks to have work! I've push an update using `this.options.lspClient.logMessage` and removed all the test setup - much better 😄

Issue #631, opened 2022-11-22
missed ?.  and filterText


jadestrong:
steps to reproduce:
```typescript
interface Test {
    name?: string;
}

const t: Test = {};

t.name.
```

```json
[Trace - 09:12:59 PM] Sending request 'completionItem/resolve - (2105)'.
Params: {
  "label": "charAt",
  "kind": 2,
  "sortText": "11",
  "commitCharacters": [
    ".",
    ",",
    "("
  ],
  "data": {
    "file": "/Users/jadestrong/Documents/Github/typescript-language-server/src/completion.ts",
    "line": 43,
    "offset": 11,
    "entryNames": [
      "charAt"
    ]
  },
  "insertTextFormat": 2,
  "textEdit": {
    "range": {
      "start": {
        "line": 42,
        "character": 9
      },
      "end": {
        "line": 42,
        "character": 10
      }
    },
    "newText": "?.charAt"
  },
  "_emacsStartPoint": 1384
}


[Trace - 09:12:59 PM] Received response 'completionItem/resolve - (2105)' in 9ms.
Result: {
  "label": "charAt",
  "kind": 2,
  "sortText": "11",
  "commitCharacters": [
    ".",
    ",",
    "("
  ],
  "data": {
    "file": "/Users/jadestrong/Documents/Github/typescript-language-server/src/completion.ts",
    "line": 43,
    "offset": 11,
    "entryNames": [
      "charAt"
    ]
  },
  "insertTextFormat": 2,
  "textEdit": {
    "range": {
      "start": {
        "line": 42,
        "character": 9
      },
      "end": {
        "line": 42,
        "character": 10
      }
    },
    "newText": "charAt(${1:pos})$0"
  },
  "_emacsStartPoint": 1384,
  "detail": "(method) String.charAt(pos: number): string",
  "documentation": {
    "kind": "markdown",
    "value": "Returns the character at the specified index.\n\n*@param* `pos` — The zero-based index of the desired character."
  },
  "insertText": "charAt(${1:pos})$0"
}
```

![image](https://user-images.githubusercontent.com/18475217/203351860-5efec781-48bf-480d-8bd8-eabc15a24f9d.png)

Should the logic of `isMemberCompletion` and `dotAccessorContext` be added to return the filterText property?

[...]

PR #678, opened 2023-02-05
feat: provide filterText property in completions

jadestrong:
fix(#631)
The implementation method refers to the implementation of vscode and volar:

vscode: https://github.com/microsoft/vscode/blob/6c866f92d7646c31bd9286b1cd8b8151c1e5b38e/extensions/typescript-language-features/src/languageFeatures/completions.ts#L100
volar(vue): https://github.com/volarjs/plugins/blob/master/packages/typescript/src/services/completions/basic.ts

### DIFF
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -159,7 +159,7 @@ function getRangeFromReplacementSpan(
         const range = ensureRangeIsOnSingleLine(optionalReplacementRange, document);
         return {
             insert: lsp.Range.create(range.start, position),
-            replace: ensureRangeIsOnSingleLine(range, document),
+            replace: range,
         };
     }
 }
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -30,6 +30,8 @@ interface CompletionContext {
         text: string;
     };
     readonly optionalReplacementSpan?: ts.server.protocol.TextSpan;
+    line: string;
+    wordRange: lsp.Range | undefined;
 }
 
 export function asCompletionItem(
@@ -62,8 +64,6 @@ export function asCompletionItem(
         },
     };
 
-    const line = document.getLine(position.line);
-
     if (entry.source && entry.hasAction) {
         // De-prioritze auto-imports
         // https://github.com/Microsoft/vscode/issues/40311
@@ -81,22 +81,22 @@ export function asCompletionItem(
     if (sourceDisplay) {
         item.detail = Previewer.plainWithLinks(sourceDisplay, filePathConverter);
     }
-    const wordRange = completionContext.optionalReplacementSpan ? Range.fromTextSpan(completionContext.optionalReplacementSpan) : undefined;
+    const { line, wordRange, isMemberCompletion, dotAccessorContext } = completionContext;
     let range: lsp.Range | ReturnType<typeof getRangeFromReplacementSpan> = getRangeFromReplacementSpan(entry, document);
     item.insertText = entry.insertText;
     item.filterText = getFilterText(entry, wordRange, line, item.insertText);
 
-    if (completionContext.isMemberCompletion && completionContext.dotAccessorContext && !entry.isSnippet) {
-        item.filterText = completionContext.dotAccessorContext.text + (item.insertText || entry.name);
+    if (isMemberCompletion && dotAccessorContext && !entry.isSnippet) {
+        item.filterText = dotAccessorContext.text + (item.insertText || entry.name);
         if (!range) {
             const replacementRange = wordRange;
             if (replacementRange) {
                 range = {
-                    inserting: completionContext.dotAccessorContext.range,
-                    replacing: rangeUnion(completionContext.dotAccessorContext.range, replacementRange),
+                    inserting: dotAccessorContext.range,
+                    replacing: rangeUnion(dotAccessorContext.range, replacementRange),
                 };
             } else {
-                range = completionContext.dotAccessorContext.range;
+                range = dotAccessorContext.range;
             }
             item.insertText = item.filterText;
         }
--- a/src/completion.ts
+++ b/src/completion.ts
@@ -88,7 +88,7 @@ export function asCompletionItem(
     if (isMemberCompletion && dotAccessorContext && !entry.isSnippet) {
         item.filterText = dotAccessorContext.text + (insertText || entry.name);
         if (!range) {
-            range = { replacing: dotAccessorContext.range };
+            range = { replace: dotAccessorContext.range };
             insertText = item.filterText;
         }
     }
@@ -124,9 +124,9 @@ export function asCompletionItem(
     }
 
     if (range) {
-        item.textEdit = range.inserting
-            ? lsp.InsertReplaceEdit.create(insertText || item.label, range.inserting, range.replacing)
-            : lsp.TextEdit.replace(range.replacing, insertText || item.label);
+        item.textEdit = range.insert
+            ? lsp.InsertReplaceEdit.create(insertText || item.label, range.insert, range.replace)
+            : lsp.TextEdit.replace(range.replace, insertText || item.label);
     } else {
         item.insertText = insertText;
     }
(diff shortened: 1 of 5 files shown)


src/completion.ts
```diff
@@ -53,13 +62,15 @@ export function asCompletionItem(
         },
     };
 
+    const line = document.getLine(position.line);
```
rchl, line 65: This is fairly expensive and is being done for every single completion while it could be done once, earlier since the result is not completion-specific.

src/lsp-server.spec.ts
```diff
@@ -274,7 +274,19 @@ describe('completion', () => {
...
+            range: {
+                start: {
+                    character: 20,
+                    line: 6,
+                },
+                end: {
+                    character: 23,
+                    line: 6,
+                },
+            },
```
rchl, line 288: Technically not wrong but it could use the "old school" insertText instead since that has smaller payload and should work the same in all cases I know (unless you know specific cases where TextEdit produces better result).

We have to consider the payload size because typescript can sometimes return 50MB worth of completions so it's worth to be as optimal as possible.

rchl (commented):
This change affects how Sublime Text presents snippet completions:

Before:

Screenshot 2023-02-05 at 16 20 46

After:

Screenshot 2023-02-05 at 16 15 12

I believe this is due to either `filterText` or `insertText` field.

I will look more into this later to see what is happening exactly but I can say right now that ST has some heuristics to decide which field to show in the completions list as it doesn't have separate concepts for "filter text" and "label". It can only filter against what is shown to the user. So it tries best to show whatever makes most sense.

Since this server is not really ST-specific, I guess the options here will be to either tweak ST heuristics to make this work better or revert some of the changes here to work more like before. I will have to understand fully what is happening and how this can affect other editors first...

src/completion.ts
```diff
@@ -70,13 +81,32 @@ export function asCompletionItem(
     if (sourceDisplay) {
         item.detail = Previewer.plainWithLinks(sourceDisplay, filePathConverter);
     }
+    const wordRange = completionContext.optionalReplacementSpan ? Range.fromTextSpan(completionContext.optionalReplacementSpan) : undefined;
```
rchl, line 84: Let's create the range earlier, where context is created, to avoid creating range for every completion.

src/completion.ts
```diff
@@ -23,14 +23,23 @@ interface ParameterListParts {
     readonly hasOptionalParameters: boolean;
 }
 
+interface CompletionContext {
+    readonly isMemberCompletion?: boolean;
```
rchl, line 27: Let's make it non-optional since the source is non-optional.

src/completion.ts
```diff
@@ -442,3 +506,9 @@ export function getCompletionTriggerCharacter(character: string | undefined): ts
             return undefined;
     }
 }
+
+function rangeUnion(a: lsp.Range, b: lsp.Range): lsp.Range {
+    const start = a.start.line < b.start.line || a.start.line === b.start.line && a.start.character < b.start.character ? a.start : b.start;
+    const end = a.end.line > b.end.line || a.end.line === b.end.line && a.end.character > b.end.character ? a.end : b.end;
+    return { start, end };
+}
```
rchl, line 514: We can move it to `src/utils/typeConverters.ts`. It already has `intersection` function.

src/lsp-server.spec.ts
rchl: Also related I guess... For import completions we have both `textEdit` and `insertText / insertTextFormat` where the latter ones should be redundant and just make the payload bigger. For example:

```json
    {
      "commitCharacters": [
        ".",
        ",",
        "("
      ],
      "data": {
        "entryNames": [
          {
            "data": {
              "exportName": "readConfigFile",
              "fileName": "/usr/local/workspace/github/typescript-language-server/node_modules/typescript/lib/typescript.d.ts",
              "moduleSpecifier": "typescript"
            },
            "name": "readConfigFile",
            "source": "typescript"
          }
        ],
        "file": "/usr/local/workspace/github/typescript-language-server/src/completion.ts",
        "line": 8,
        "offset": 17
      },
      "detail": "typescript",
      "filterText": "import { readConfigFile$1 } from 'typescript';",
      "insertText": "import { readConfigFile$1 } from 'typescript';",
      "insertTextFormat": 2,
      "kind": 3,
      "label": "readConfigFile",
      "sortText": "11",
      "textEdit": {
        "insert": {
          "end": {
            "character": 18,
            "line": 7
          },
          "start": {
            "character": 0,
            "line": 7
          }
        },
        "newText": "import { readConfigFile$1 } from 'typescript';",
        "replace": {
          "end": {
            "character": 18,
            "line": 7
          },
          "start": {
            "character": 0,
            "line": 7
          }
        }
      }
    }
```

src/completion.ts
```diff
@@ -23,14 +23,25 @@ interface ParameterListParts {
     readonly hasOptionalParameters: boolean;
 }
 
+interface CompletionContext {
+    readonly isMemberCompletion?: boolean;
+    readonly dotAccessorContext?: {
+        range: lsp.Range;
+        text: string;
+    };
+    readonly optionalReplacementSpan?: ts.server.protocol.TextSpan;
```
rchl, line 32: We don't need this anymore?

src/completion.ts
```diff
@@ -23,14 +23,25 @@ interface ParameterListParts {
...
+    line: string;
+    wordRange: lsp.Range | undefined;
```
rchl, line 34: Can we call it `optionalReplacementRange`?

src/completion.ts
```diff
@@ -81,15 +80,15 @@ export function asCompletionItem(
...
     let range: lsp.Range | ReturnType<typeof getRangeFromReplacementSpan> = getRangeFromReplacementSpan(entry, document);
     item.insertText = entry.insertText;
-    item.filterText = getFilterText(entry, wordRange, line, item.insertText);
+    item.filterText = getFilterText(entry, optionalReplacementRange, line, item.insertText);
 
     if (isMemberCompletion && dotAccessorContext && !entry.isSnippet) {
         item.filterText = dotAccessorContext.text + (item.insertText || entry.name);
         if (!range) {
-            const replacementRange = wordRange;
+            const replacementRange = optionalReplacementRange;
```
rchl, line 91: This variable is not necessary.

src/lsp-server.spec.ts
rchl: OK, I won't insist on using `insertText` over `textEdit` since that is inferior and can in theory produce worse results (when part of the prefix includes trigger (like punctuation) characters.

But we can instead get rid of `insertText` (not set it) when `textEdit` is set.

rchl:
> Since this server is not really ST-specific, I guess the options here will be to either tweak ST heuristics to make this work better or revert some of the changes here to work more like before. I will have to understand fully what is happening and how this can affect other editors first...

I will most likely have to address this in ST somehow since it's nothing particularly wrong with this change.

jadestrong:
The current implementation will give priority to InsertReplaceEdit. I haven't figured out how to give priority to TextEdit.

src/completion.ts
```diff
@@ -101,44 +129,80 @@ export function asCompletionItem(
...
+
+    if (replacementSpan && replacementSpan.start.line !== replacementSpan.end.line) {
+        replacementSpan = lsp.Range.create(replacementSpan.start, document.getLineEnd(replacementSpan.start.line));
     }
-    return range;
+
+    return {
+        inserting: replacementSpan,
+        replacing: replacementSpan,
+    };
```
rchl, line 205: The previous code has a comment:

```ts
// If TS provides an explicit replacement span with an entry, we should use it and not provide an insert.
```

and also only provided a `textEdit` in this case since it doesn't really make sense to provide both insert & replace if both are the same anyway.

Can we keep this logic from before?

src/completion.ts
```diff
@@ -101,44 +129,80 @@ export function asCompletionItem(
...
+    }
+
+    // In all other cases, fallback to using the insertText
+    return insertText;
 }
 
-function ensureRangeIsOnSingleLine(range: lsp.Range, document: LspDocument): lsp.Range {
-    if (range.start.line !== range.end.line) {
-        return lsp.Range.create(range.start, document.getLineEnd(range.start.line));
+function getRangeFromReplacementSpan(entry: ts.server.protocol.CompletionEntry, document: LspDocument) {
```
rchl, line 191: I'd like return types to be explicitly specified.

src/lsp-server.ts
```diff
@@ -625,13 +625,32 @@ export class LspServer {
...
             for (const entry of entries || []) {
                 if (entry.kind === 'warning') {
                     continue;
                 }
-                const completion = asCompletionItem(entry, optionalReplacementSpan, file, params.position, document, this.documents, completionOptions, this.features);
+                const completion = asCompletionItem(entry, file, params.position, document, this.documents, completionOptions, this.features, {
+                    isMemberCompletion,
+                    dotAccessorContext,
+                    line: document.getLine(params.position.line),
+                    optionalReplacementRange: optionalReplacementSpan ? Range.fromTextSpan(optionalReplacementSpan) : undefined,
```
rchl, line 652: This is still done for each completion while it only needs to be done one.

rchl: addressed

rchl (approved):

rchl:
Thanks!

PR #685, opened 2023-02-11
feat: support `workspace/willRenameFiles` request

antosha417:
Hey! I've been working on a plugin for neovim that automatically applies workspace edits before moving files around. And it does not work with typescript. The problem with `_typescript.applyRenameFile` that edit is applied only after neovim moves the file. 

So I wanted to implement `workspase/willRenameFile` command that returns the edit. We will be then able to preview and apply it if needed.

Closes https://github.com/typescript-language-server/typescript-language-server/issues/665
Closes https://github.com/antosha417/nvim-lsp-file-operations/issues/6

### DIFF
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -256,6 +256,13 @@ export class LspServer {
                     full: true,
                     range: true,
                 },
+                workspace: {
+                    fileOperations: {
+                        willRename: {
+                            filters: [{ pattern: { glob: '**/*.{ts,js}', matches: 'file' } }],
+                        },
+                    },
+                },
             },
         };
         if (textDocument?.callHierarchy && typescriptVersion.version?.gte(API.v380)) {
@@ -1075,6 +1082,32 @@ export class LspServer {
             edit: { changes },
         });
         return applied;
+    }
+
+    async willRenameFiles(params: lsp.RenameFilesParams, token?: lsp.CancellationToken): Promise<lsp.WorkspaceEdit> {
+        const toTextEdit = (codeEdit: ts.server.protocol.CodeEdit): lsp.TextEdit => {
+            return {
+                range: {
+                    start: Position.fromLocation(codeEdit.start),
+                    end: Position.fromLocation(codeEdit.end),
+                },
+                newText: codeEdit.newText,
+            };
+        };
+
+        const workspaceEdit: lsp.WorkspaceEdit = { changes: {} };
+        for (const rename of params.files) {
+            const codeEdits = await this.getEditsForFileRename(rename.oldUri, rename.newUri, token);
+            for (const codeEdit of codeEdits) {
+                const uri = pathToUri(codeEdit.fileName, this.documents);
+                if (!workspaceEdit.changes) {
+                    workspaceEdit.changes = {};
+                }
+                const textEdits = workspaceEdit.changes[uri] || (workspaceEdit.changes[uri] = []);
+                textEdits.push(...codeEdit.textChanges.map(toTextEdit));
+            }
+        }
+        return workspaceEdit;
     }
 
     protected async applyRenameFile(sourceUri: string, targetUri: string, token?: lsp.CancellationToken): Promise<void> {
--- a/src/lsp-server.ts
+++ b/src/lsp-server.ts
@@ -261,7 +261,7 @@ export class LspServer {
                         willRename: {
                             filters: [{
                                 scheme: 'file',
-                                pattern: { glob: '**/*.{ts,js,jsx,tsx,mjs,mts,cjs}', matches: 'file' },
+                                pattern: { glob: '**/*.{ts,js,jsx,tsx,mjs,mts,cjs,cts}', matches: 'file' },
                             }],
                         },
                     },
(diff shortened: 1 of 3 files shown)


src/lsp-server.ts
```diff
@@ -256,6 +256,13 @@ export class LspServer {
                     full: true,
                     range: true,
                 },
+                workspace: {
+                    fileOperations: {
+                        willRename: {
+                            filters: [{ pattern: { glob: '**/*.{ts,js}', matches: 'file' } }],
```
predragnikolic, line 262: do we also need:
jsx
tsx
mjs
mts
cjs
cts

.d.ts
.d.mts
.d.cts

in the globs?

predragnikolic: Just let me check if I accidentally didn't made up some of the above file extensions.

rchl, line 262: Could also add `scheme: 'file'` since the server doesn't currently work for other unsaved files.

src/lsp-server.ts
```diff
@@ -1077,6 +1084,32 @@ export class LspServer {
...
+    async willRenameFiles(params: lsp.RenameFilesParams, token?: lsp.CancellationToken): Promise<lsp.WorkspaceEdit> {
+        const toTextEdit = (codeEdit: ts.server.protocol.CodeEdit): lsp.TextEdit => {
+            return {
+                range: {
+                    start: Position.fromLocation(codeEdit.start),
+                    end: Position.fromLocation(codeEdit.end),
+                },
+                newText: codeEdit.newText,
+            };
+        };
```
rchl, line 1096: Such function already exists: https://github.com/typescript-language-server/typescript-language-server/blob/2c9eb632659a3bb9995095576afe88e84833bbdd/src/protocol-translation.ts#L171-L171

src/lsp-server.ts
```diff
@@ -1077,6 +1084,32 @@ export class LspServer {
...
+            for (const codeEdit of codeEdits) {
+                const uri = pathToUri(codeEdit.fileName, this.documents);
+                if (!workspaceEdit.changes) {
+                    workspaceEdit.changes = {};
+                }
+                const textEdits = workspaceEdit.changes[uri] || (workspaceEdit.changes[uri] = []);
+                textEdits.push(...codeEdit.textChanges.map(toTextEdit));
+            }
+        }
+        return workspaceEdit;
```
rchl, line 1110: Just a style thing but this is shorter and cleaner:

```suggestion
        const changes: lsp.WorkspaceEdit['changes'] = {};
        for (const rename of params.files) {
            const codeEdits = await this.getEditsForFileRename(rename.oldUri, rename.newUri, token);
            for (const codeEdit of codeEdits) {
                const uri = pathToUri(codeEdit.fileName, this.documents);
                const textEdits = changes[uri] || (changes[uri] = []);
                textEdits.push(...codeEdit.textChanges.map(toTextEdit));
            }
        }
        return { changes };
```

src/lsp-server.ts
rchl: I don't think `cts` is a thing.

And the `.d.*` variants would probably be covered by `.*`? Not sure of the glob behavior here.

antosha417: I'm not sure what `scheme` means. I see this in spec:
```
	/**
	 * A Uri like `file` or `untitled`.
	 */
	scheme?: string;
```

rchl: `scheme: file` would act as an additional filter so that the request should not triggered for files that are not saved on disk for example since server can't handle those.

src/lsp-server.ts
antosha417: I looked at `rename` function. Will fix there also.

src/lsp-server.ts
antosha417: > And the .d.* variants would probably be covered by .*?

yes `filename.d.ts` will match `*.ts` glob

rchl (approved):

src/lsp-server.ts
predragnikolic: https://www.typescriptlang.org/docs/handbook/esm-node.html

> In turn, TypeScript supports two new source file extensions: .mts and .cts. When TypeScript emits these to JavaScript files, it will emit them to .mjs and .cjs respectively.

src/lsp-server.ts
```diff
@@ -256,6 +256,16 @@ export class LspServer {
                     full: true,
                     range: true,
                 },
+                workspace: {
+                    fileOperations: {
+                        willRename: {
+                            filters: [{
+                                scheme: 'file',
+                                pattern: { glob: '**/*.{ts,js,jsx,tsx,mjs,mts,cjs}', matches: 'file' },
```
predragnikolic, line 264:
```suggestion
                                pattern: { glob: '**/*.{ts,js,jsx,tsx,mjs,mts,cjs,cts}', matches: 'file' },
```

predragnikolic (approved):
Cool stuff!

This means that the typescript language server can deprecate the custom `_typescript.applyRenameFile` request at some point

rchl:
Thanks

PR #688, opened 2023-02-15
feat: start separate tsServer instance for semantic requests

### LINKED ISSUE #650: [Request for info] open performance work
Hello team, first of all thanks for putting all that work in typescript-language-server.

I'm opening this issue as a "request for information". Tldr: could you describe all open performance topics so contributors can easily pick topics up and submit PRs?

AFAIK, since day and age, this LSP gives a slower experience than the non-LSP compliant tsserver experience you get with VsCode. Microsoft does not want to provide a LSP spec complient lang server (https://github.com/microsoft/TypeScript/issues/39459; my free interpretation of their lack of progress here). I personally use neovim, and combined with the nvim-cmp autocompletion plugin, in many codebases the performance is fine for me on my M1 Mac, but some others I notice the difference in performance. For example, try adding the `aws-sdk` package to your project and see how slow autocompletion gets (even VsCode will get slower; but performs better).

I remember neovim contributers say typescript-language-server is one of the slowest out there, something with returning too many results, and not being the fault of the authors because it's also due to working around tsserver, but not sure if that is still true.

If I can, I'd love to see if I can contribute, because I don't want to switch editor for codebases that get slow with the lang server. 

I imagine having a set of well described issues with a "perf" label of sorts would help jumping in!

### LINKED ISSUE #675: Recognize new import takes too long in large repos
Hi folks.

__TLDR:__ The server is taking more than 1 minute to recognize new exports on modules, and only after the file that is making the export is saved. This happens in both Neovim and [Helix](https://helix-editor.com/), so it seems that isn't an issue with a specific configuration.

Neovim config (just in case):
``` lua
local nvim_lsp = require('lspconfig')

nvim_lsp.tsserver.setup{
  on_attach = on_attach,
  capabilities = require('cmp_nvim_lsp').default_capabilities(),
  flags = {
    debounce_text_changes = 150,
  },
  maxTsServerMemory = 4096,
  init_options = {
    preferences = {
      disableSuggestions = true,
    }
  }
}
```

__Neovim version:__ 0.8.1
__Node version:__ 16.10.0
__typescript-language-server version:__ 3.0.3
__typescript version:__ 4.9.4

__More context:__

In the company I work for there's a Typescript repository that's growing very fast. They migrated recently to [Nx](https://nx.dev/) build system, and the repository is quite large at the moment. A few weeks ago I updated some of the plugins and language servers in my Neovim configuration and I started noticing that in this repo there was quite a huge delay in some of the typescript-language-server diagnostics, especially when adding a new export in a module, and then trying to import it from other module.

If I don't save the file where the export is added, and try to import the new item in another file, then the server never recognizes that the module is exporting that new item, and the `Module 'example/file' has no exported member 'newItem'` error never hides. That wasn't happening before the update.

Then, if I save the file, the error `Module 'example/file' has no exported member 'newItem'` in the file that is importing is only hidden after a minute and half. This means that the server takes more than a minute to recognize the new export, which is clearly too long.

I tried this in both Neovim and Helix, with the same result (even after adding the `maxTsServerMemory` config). I then downloaded VSCode and did the same test with the out of the box configuration. It recognizes the new import almost instantly and without saving the file that make the export.

Am I missing a configuration in the server to improve this in large repos?

rchl:
Ported code from VSCode that enables two tsserver instances, one for "syntax" requests and one for "semantic" requests. This should ensure that the more expensive requests like semantic tokens or diagnostic computations don't block more acute requests like completions.

The default is to run with syntax and semantic servers enabled.

Resolves #650 (?)
Resolves #675 (?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,16 @@ interface TsserverOptions {
      * @default 'off'
      */
     trace?: 'off' | 'messages' | 'verbose';
+    /**
+     * Controls if TypeScript launches a dedicated server to more quickly handle syntax related operations, such as computing diagnostics or code folding.
+     *
+     * Allowed values:
+     *  - auto: Spawn both a full server and a lighter weight server dedicated to syntax operations. The syntax server is used to speed up syntax operations and provide IntelliSense while projects are loading.
+     *  - never: Don't use a dedicated syntax server. Use a single server to handle all IntelliSense operations.
+     *
+     * @default 'auto'
+     */
+    useSyntaxServer?: 'auto' | 'never';
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ interface TsserverOptions {
      */
     trace?: 'off' | 'messages' | 'verbose';
     /**
-     * Controls if TypeScript launches a dedicated server to more quickly handle syntax related operations, such as computing diagnostics or code folding.
+     * Whether a dedicated server is launched to more quickly handle syntax related operations, such as computing diagnostics or code folding.
      *
      * Allowed values:
      *  - auto: Spawn both a full server and a lighter weight server dedicated to syntax operations. The syntax server is used to speed up syntax operations and provide IntelliSense while projects are loading.
--- a/src/configuration-manager.ts
+++ b/src/configuration-manager.ts
@@ -128,7 +128,7 @@ export class ConfigurationManager {
                 autoImportFileExcludePatterns: this.getAutoImportFileExcludePatternsPreference(workspaceFolder),
             },
         };
-        client.executeWithoutWaitingForResponse(CommandTypes.Configure, args);
+        client.request(CommandTypes.Configure, args);
     }
 
     public async configureGloballyFromDocument(filename: string, formattingOptions?: lsp.FormattingOptions): Promise<void> {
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -118,6 +118,16 @@ interface TsserverOptions {
...
+    /**
+     * Whether a dedicated server is launched to more quickly handle syntax related operations, such as computing diagnostics or code folding.
+     *
+     * Allowed values:
+     *  - auto: Spawn both a full server and a lighter weight server dedicated to syntax operations. The syntax server is used to speed up syntax operations and provide IntelliSense while projects are loading.
+     *  - never: Don't use a dedicated syntax server. Use a single server to handle all IntelliSense operations.
+     *
+     * @default 'auto'
+     */
+    useSyntaxServer?: 'auto' | 'never';
```
rchl, line 130: It looks like it should be a boolean value but it's this way because VSCode also supports `always` and I want to have this future-proof in case support for `always` is added later.

To support `always` we'd need to add a bunch of checks to avoid triggering certain requests when it's used. Also, I believe in VSCode `always` is forced by the editor when opening single files and is not really meant to be used manually. So it might required some client glue to be usable in other editors.

PR #768, opened 2023-10-11
fix(perf): cache completion data on the server to avoid serialization

rchl:
A perf improvement to avoid sending the whole completion data (the `data` property of the `CompletionItem`) to the client.

Cache completion data in a map on the server and only send a `cacheId` that references a cached entry to the client. That way the completion data can be looked up on the server during `completion/resolve` by `cacheId`.

Results in much smaller `textDocument/completion` responses (620KB vs. 200KB in an example project with `aws-sdk` installed).

Inspired by a https://github.com/yioneko/vtsls/issues/26#issuecomment-1754909555 comment (even though that's about code actions).


rchl:
BTW. It assumes that there can be only one completion request at a time (new one invalidates the previous one). Not sure if that holds 100% of the time but should be fine in practice.

PR #782, opened 2023-10-29
refactor: port ITypeScriptServiceClient

rchl:
This is a refactor to bring the code closer to the original VSCode implementation and make things easier to port in the future.

This should mostly have no observable differences besides some intentional smaller changes:
 - updated default settings used for projects that don't include `tsconfig`. New defaults include
	 - `moduleResolution: 'bundler'` (if TS v5+)
	 - `target: 'es2022'`
	 - `allowImportingTsExtensions: true` (if TS v5+)
	 - default to unix newlines when we don't know used newlines (LSP only reports newline character in formatting actions)
 - dropped some special cases for TS v2.50 and lower (no longer supported)
 - added handling for canceling some in-flight requests when document changes


rchl:
I'm disabling minification of built bundle since I found that otherwise, if the server crashes and reports an error, the whole minified line will be printed in the error message which hides the real error message and stack trace.

Also, it's easier to audit changes if built code is not aggressively minifed.

PR #785, opened 2023-11-01
feat: support code lens for references and implementations

### LINKED ISSUE #5: CodeLens support
Like I had suggested at https://github.com/prabirshrestha/typescript-language-server/issues/9

tsserver doesn't supports CodeLens, but we could implement `textDocument/codeLens` and `codeLens/resolve` like VSCode does for [References CodeLens](https://github.com/Microsoft/vscode/blob/master/extensions/typescript/src/features/referencesCodeLensProvider.ts) and [Implementations CodeLens](https://github.com/Microsoft/vscode/blob/master/extensions/typescript/src/features/implementationsCodeLensProvider.ts)

rchl:
Ported Code Lens functionality from VSCode.

Resolves #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,11 @@ Some of the preferences can be controlled through the `workspace/didChangeConfig
 [language].inlayHints.includeInlayPropertyDeclarationTypeHints: boolean;
 [language].inlayHints.includeInlayVariableTypeHints: boolean;
 [language].inlayHints.includeInlayVariableTypeHintsWhenTypeMatchesName: boolean;
+// Code Lens preferences
+[language].implementationsCodeLens.enabled: boolean;
+[language].referencesCodeLens.enabled: boolean;
+[language].referencesCodeLens.showOnAllFunctions: boolean;
+
 /**
  * Complete functions with their parameter signature.
  *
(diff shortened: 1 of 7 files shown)


predrag-codetribe:
Nice!

Implementation count doesn't seem correct, with the following code:
```
interface Int {}

class Hello implements Int {}
class Hello2 implements Int {}
```
VS code shows the correct count.

Screenshot 2023-11-02 at 11 20 25

rchl:
Very nice catch. It worked fine, just that the title was wrong.

(planning to add some basic tests for this so I would hopefully catch it also ;))

predrag-codetribe:
I have `implementationsCodeLens.enabled` and `typescript.referencesCodeLens.enabled` disabled by default:
```
{
  "settings": {
		"typescript.implementationsCodeLens.enabled": false,
		"typescript.referencesCodeLens.enabled": false,
}
```

But I still see these lenses by default.

rchl:
Right. Seems like I've completely overlooked that part.

predragnikolic:
There is also one more setting to support... `referencesCodeLens.showOnAllFunctions` 🙃

rchl:
It is supported. You will see it referenced in the changes of you search for "showOnAllFunctions"

PR #814, opened 2023-11-16
fix: add folder filter in server capabilities for `willRename`

luckasRanarison:
typescript-language-server does support willRenameFiles on `folder` but it only provides filter for `file`, other servers like rust-analyzer explicitly specify this filter. The absence of this filter also causes issues in willRenameFiles implementation on the client side if the client only tracks things specified by the filters, i.e it doesn't send requests on folder renamed.

### DIFF
--- a/src/lsp-server.spec.ts
+++ b/src/lsp-server.spec.ts
@@ -2385,12 +2385,14 @@ describe('completions without client snippet support', () => {
 });
 
 describe('fileOperations', () => {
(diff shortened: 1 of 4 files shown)


rchl:
Please cherry pick d9c102d32b99fe8cea1a0a7da32b2ed34ccda4e7 from this repo. I've added tests and fixed linting issues.

rchl (approved):

rchl:
Thanks!

PR #851, opened 2024-01-12
fix: out of date diagnostics after closing and re-opening a file

rchl:
diagnostics would malfunction for a file that was closed and later re-opened

### DIFF
--- a/src/diagnostic-queue.ts
+++ b/src/diagnostic-queue.ts
@@ -50,8 +50,8 @@ class FileDiagnostics {
     }
 
     public onDidClose(): void {
-        this.publishDiagnostics();
         this.diagnosticsPerKind.clear();
+        this.publishDiagnostics();
         this.closed = true;
     }
 
--- a/src/diagnostic-queue.ts
+++ b/src/diagnostic-queue.ts
@@ -106,6 +106,7 @@ export class DiagnosticEventQueue {
         const uri = this.client.toResource(file).toString();
         const diagnosticsForFile = this.diagnostics.get(uri);
         diagnosticsForFile?.onDidClose();
+        this.diagnostics.delete(uri);
     }
 
     /**


src/diagnostic-queue.ts
```diff
@@ -106,6 +106,7 @@ export class DiagnosticEventQueue {
         const uri = this.client.toResource(file).toString();
         const diagnosticsForFile = this.diagnostics.get(uri);
         diagnosticsForFile?.onDidClose();
+        this.diagnostics.delete(uri);
```
predrag-codetribe, line 109: This PR does fix the issue that I have experienced. Thanks.

---

One question I have is if you want to match the VS Code behavior?
 
I compared the behavior of VS Code and this PR.

VS code will clear the diagnostic as soon as file_a is closed.

In this PR, 
the diagnostics will not be cleared until the file_a is open again.

Which makes me think if the diagnostics should be published to the client on did close, 
if you want to match with VS Code behavior?

rchl: yeah, the intention of `onDidClose` was to clear diagnostics before closing. should be fixed now?

predragnikolic: I would not move the call of this.publishDiagnostics() to a different place. The previous position was ok.

But I would swap 109 line with 108,
So that `this.diagnostics.delete(uri);` is executed first before onDidClose is called.

Wdyt?

rchl: I'm not sure how would that help. Wouldn't `FileDiagnostics` instance still publish diagnostics that were reported at the time of closing the file?

(I'm looking at this while trying to do real work so I might be a bit distracted)

predragnikolic: > (I'm looking at this while trying to do real work so I might be a bit distracted)

Same, and I am even writing the review on my phone. I will leave this out for later today... :⁠-⁠)

predragnikolic: I was wrong.

predragnikolic (approved):
Works great!

predragnikolic:
I just read the LSP spec, 
and I think that we should just be aware what the spec says:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics
> - if a language is single file only (for example HTML) then diagnostics are cleared by the server when the file is closed. Please note that open / close events don’t necessarily reflect what the user sees in the user interface. These events are ownership events. So with the current version of the specification it is possible that problems are not cleared although the file is not visible in the user interface since the client has not closed the file yet.
> - if a language has a project system (for example C#) diagnostics are not cleared when a file closes. When a project is opened all diagnostics for all files are recomputed (or read from a cache).

TS server belongs to the list of languages that has projects, so I think this PR might affect to sometimes clear diagnostics when it shouldn't. 

As I said, I want us to be aware of what the spec says.
This PR brings the behavior of how VS Code behaves and fixes one issue.

rchl:
This server doesn't provide diagnostics for all project files so the second case doesn't apply to it and it would be wrong to not clear file diagnostics on closing the file then.

PR #918, opened 2024-09-27
fix(previewer): mark `@example` and `@default` code blocks as TS

ribru17:
When interpreting the preview result as Markdown, it is quite helpful to have the `typescript` language annotation before `@example` code blocks because it will give code highlights and make it more clear that this is an example of code rather than just raw text.


rchl (approved):

rchl (approved):
Nothing really guarantees that it will be a typescript snippet but I guess it might be in majority of cases.

PR #953, opened 2025-04-18
Tell client a completion item has code actions

wyuenho:
This will allow clients to better detect which completion item really needs to be resolved. This will help speed up clients such as Emacs' lsp-mode

https://github.com/emacs-lsp/lsp-mode/issues/4757


wyuenho:
Sorry, turns out this is not needed

PR #1027, opened 2025-10-17
feat: Add support for diagnosticsIgnoredCodes in initialize.

ShogunPanda:
This PR adds support for `diagnosticsIgnoredCodes` in the initialize options so that LSP clients that don't send `workspace/didChangeConfiguration` can set the value.


rchl:
I suggest for pushing implementation in the client. This is just not right. It's not the responsibility of the server to cater for client deficits.

PR #1057, opened 2025-11-21
feat: Add support for client-side file watcher

### LINKED ISSUE #956: Pass `--canUseWatchEvents` to tsserver
Is there a way to set the `--canUseWatchEvents` flag when [starting tsserver](https://github.com/microsoft/TypeScript/blob/main/src/tsserver/server.ts#L48)? I've tried passing it as a command line arg to typescript-language-server in my Neovim LSP config:
```
vim.lsp.config('ts_ls', {
  cmd = { "typescript-language-server", "--stdio", "--canUseWatchEvents" },
})
```
... but it errors with "error: unknown option '--canUseWatchEvents'".

For background, typescript-language-server appears to be actively watching 40k files in my large project. I noticed that VS Code in the same project only watches around 200 files. I've read that VS Code uses a different version of node (electron-based?) with pointer compression so that could account for some improvements. That being said, I noticed that VS Code starts tsserver with the `--canUseWatchEvents` flag and typescript-language-server does not so that seemed like a good starting point.

Is there an option or workaround that I can use to test this theory?

OleksandraKordonets:
Fixes #956
This PR adds optional support for a new TypeScript server feature that allows the editor to provide file watch events instead of having tsserver watch all files directly. When enabled and supported, this can significantly reduce the number of filesystem watchers in large workspaces.

Changes:
- Added a new optional CLI option: `--canUseWatchEvents`
- Passed `--canUseWatchEvents` to tsserver when supported
- Implemented a `WatchEventManager` to track tsserver watcher requests and manage dynamic workspace watchers
- Forwarded relevant client file change notifications to tsserver using the `watchChange` request
- Added support for new tsserver events: `createFileWatcher`, `createDirectoryWatcher`, `closeFileWatcher`
- Added fallback behavior with warnings when prerequisites are not met

When this feature is enabled and all requirements are satisfied the server will:
- Launch tsserver with `--canUseWatchEvents`
- Register minimal LSP file watchers instead of tsserver watching everything
- Forward `DidChangeWatchedFiles` events to tsserver as `watchChange` notifications
- If any prerequisite is missing, the server logs an explanation and falls back to the existing behavior, ensuring there is no change or risk for existing users.

Notes:
Requires TypeScript >= 5.4 and client support for dynamic watched-files registration and relative patterns. And if any prerequisite is missing, a warning is logged and the server continues with normal tsserver file watching, so no behavior change for existing users.


rchl:
I have actually been working on it last week and came quite far. One thing that was a bit problematic was to adapt the watcher ID concept into LSP where there are no IDs. Curious to check how you've handled that.

After I'll check the changes out we'll have to decide which changes to go with based on what would be more efficient.

rchl:
After cursory look I have some comments but I think it would be better to just go with my version of the code which is based on vscode's implementation more closely as that makes maintenance easier.

Some notes (but not sure you should spend time on those given what I said above):
- [x] The setting whether to enable events should be passed through initializationOptions, not cli.
- [x] Server explicitly doesn't support workspace folders (has no corresponding server capability) so there is no need to try to read those.
- [ ] Event manager doesn't do debouncing when notifying tsserver about events. The original vscode code does.
- [x] The ts events related to watching are not currently handled so basically all the relevant code doesn't run
- [x] There are issues with un-registering capabilities. It unregisters same capability multiple times rather than always the last one. Think there is a race condition in setting `this.registration`.
- [x] The events are not matched properly against registered watchers. I had to do some quick hack fixes to get things running in the first place but looks like simple file change events are not reported to tsserver.
- [ ] There is a quite a lot of custom code for handling paths (normalizing, getting dirname, replacing (back)slashes). I wouldn't trust such code personally since there is a lot of corner cases that it might not handle. There are already established ways for doing that (URI class, os.path functions) that should generally work better. Assuming that it's even necessary to do all that.
- [x] The most tricky part for getting this to work is matching file events from the client to tsserver watcher IDs and I'm not sure how well it works given the problems above.

rchl:
I appreciate your efforts on getting registrations aggregated. Since tsserver is very granular in its events, I think your approach makes sense where you mainly just watch the root and match against tsserver watchers once the events are received. That said, I would still prefer to match vscode more closely in some parts.

Perhaps we can work together on this branch to get it working and finalized? That would be more efficient than going through PR comments. Alternatively I could work on my branch and potentially pick some stuff from you.

OleksandraKordonets:
> I appreciate your efforts on getting registrations aggregated. Since tsserver is very granular in its events, I think your approach makes sense where you mainly just watch the root and match against tsserver watchers once the events are received. That said, I would still prefer to match vscode more closely in some parts.
> 
> Perhaps we can work together on this branch to get it working and finalized? That would be more efficient than going through PR comments. Alternatively I could work on my branch and potentially pick some stuff from you.

Thanks so much for the detailed review and for taking the time to run and investigate my implementation. I really appreciate the explanations, a lot of these areas were things I didn’t fully understand when I started, and I learned a lot while working on this.
For me, this was mainly a learning experience and a way to understand how everything fits together, so feedback like this is really valuable. 

I’m glad some of the ideas were useful. Since you already have a more complete implementation underway, I think it makes the most sense for you to continue with your version. Please feel free to take anything from my branch that helps.

rchl:
No problem.

Lets work on this together then. I can push to the branch so it should be easy and we can discuss things as we go.

OleksandraKordonets:
@rchl 
Hi, thanks a lot for pushing those changes and continuing the work on this. I was a bit busy over the past couple of weeks, but I’m catching up now and will review them shortly.

rchl:
I also didn't have much time to work on it so far. I'm thinking now (against some of my thoughts before) to keep most of your design instead of aligning more with the source since that would take too much time.

I would say it's still not ready for serious review.

rchl:
Review from claude:

### Code Quality & Correctness

**Race condition in `updateRegistration`** — This is the most significant issue. The method is called fire-and-forget (`void this.updateRegistration()`) in multiple places. The `registrationSignature` guard is set *synchronously* before `await`, so if two concurrent calls overlap, the second can dispose `this.registration` while it's still `undefined` (the first hasn't resolved yet), then the first call resolves and overwrites the second's registration with a stale one:

```ts
// updateRegistration — concurrent call scenario:
// Call 1: sets signature=S1, does dispose (noop), awaits → gets R1
// Call 2: sets signature=S2, calls this.registration?.dispose() → still undefined!, awaits → gets R2
// Call 1 resolves: this.registration = R1  ← overwrites R2, R2 is leaked
```

A simple fix is a serial update queue or a generation counter:

```ts
private pendingRegistration: Promise<void> | undefined;

private updateRegistration(): void {
    this.pendingRegistration = (this.pendingRegistration ?? Promise.resolve()).then(() => this.doUpdateRegistration());
}
```

**`dispose` doesn't cancel in-flight `updateRegistration`** — After `dispose()` clears state and calls `this.registration?.dispose()`, an in-flight `updateRegistration` can still resolve and set `this.registration` on the now-disposed instance. Same generation-counter fix resolves this too.

**`watcherMatchesPath` — non-recursive directory case** — For a non-recursive directory watcher at `/foo`, the check `this.dirname(normalizedFsPath) === base` is correct for files inside `/foo`. But a change event for `/foo` itself (the directory) won't match because `dirname('/foo')` → `''` or `'/'`. This is likely intentional (we watch the contents, not the dir node itself), but worth a comment.

**`closeWatcher` usage counter fallback** — `(this.coverageUsage.get(watcher.coverageKey) ?? 1) - 1` — using `1` as default means if the key is missing it goes to `0` and cleans up. Using `0` would be more conservative (noop). The current behavior is probably fine but the intent isn't obvious.

---

### Test Coverage

- **No unit tests for `WatchEventManager`**. This class has non-trivial logic: coverage key deduplication, `watchKind` upgrading, recursive vs. non-recursive matching, case-insensitive path comparison, and the coverage/usage reference counting. All of these warrant direct unit tests.
- The only test change ([`src/ts-client.test.ts:41`](src/ts-client.test.ts)) adds `useClientFileWatcher: false` to an existing test — purely a compilation fix for the updated interface.
- `TestLspClient.registerDidChangeWatchedFilesCapability` throws `Error('unsupported')`. Consider adding a spy/stub version for integration tests that exercise the watch path end-to-end.

src/watchEventManager.ts
```diff
@@ -106,7 +106,6 @@ export class WatchEventManager {
             const uri = URI.parse(change.uri);
             const fsPath = uri.fsPath;
             const normalizedFsPath = this.normalizePath(fsPath);
-            const tsserverPath = this.toTsserverPath(fsPath);
```
rchl: I wonder why can't we send `normalizedFsPath` to tsserver instead of `tsserverPath`...?

Maybe I'm gambling here a bit but I've changed it to send `normalizedFsPath`.

rchl (approved):

rchl:
@OleksandraKordonets it's been a while since we've looked at it, I guess, but I think it's in a state that I'd be OK shipping.

PR #1064, opened 2026-03-04
feat: add opt-in eager diagnostic clearing on didChange

alliprice:
## Summary

Adds an opt-in `diagnostics.eagerClear` setting that publishes empty diagnostics for a file immediately when its content changes, before reanalysis completes.

## Motivation

LSP servers are optimized for interactive editors, where debouncing diagnostics reduces flicker while a human is typing. This default is sensible and should remain unchanged.

However, LSP is increasingly used by headless agentic tools (AI coding assistants, automated refactoring systems) that consume diagnostics programmatically. For these clients, stale diagnostics between an edit and reanalysis create incorrect input to decision loops, causing issues like:

- Re-fixing already fixed errors
- Missing newly introduced errors
- Looping on diagnostics that no longer apply

This is not a bug in the server - it is a difference in client requirements. Interactive editors benefit from stable UI during typing. Programmatic clients benefit from immediate staleness signals.

## Changes

- **New setting**: `diagnostics.eagerClear` (default: `false`)
  - When enabled, `textDocument/didChange` triggers an immediate publish of empty diagnostics for the changed file
  - Server continues normal debounced analysis pipeline
  - Fresh diagnostics arrive once tsserver reanalysis completes

- **Implementation**:
  - Adds `clearDiagnosticsForFile()` method to `DiagnosticsManager`
  - Reuses existing `onDidClose()` invalidation path to cancel pending debounced publishes before clearing
  - Hooks into `didChangeTextDocument` handler when setting is enabled

- **Tests**: Two new tests verify:
  - When enabled: diagnostics are cleared synchronously on `didChange`
  - When disabled: existing behavior preserved (no synchronous clear)

## Configuration

```json
{
  "diagnostics.eagerClear": false
}
```

Setting this to `true` is intended for non-UI rapid-edit workflows such as agentic coding tools.

## Test Plan

1. Run existing test suite to verify no regressions in default behavior
2. Run new tests (`yarn test:commit`) to verify eager clearing works when enabled
3. Manual verification in VS Code:
   - With setting disabled: observe normal debounced diagnostic updates
   - With setting enabled: observe immediate diagnostic clearing on edit

## Related Work

This is part of a coordinated effort across multiple LSP server ecosystems to support agentic workflows. See: https://github.com/lsp-eager-invalidation

Similar PRs:
- pyright (Python)
- rust-analyzer (Rust)
- gopls (Go)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -227,6 +227,15 @@ completions.completeFunctionCalls: boolean;
 // See https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json for a full list of valid codes.
 diagnostics.ignoredCodes: number[];
 /**
+ * When enabled, diagnostics for a file are cleared immediately on each change,
+ * before fresh diagnostics arrive from tsserver. This prevents stale diagnostics
+ * from persisting between an edit and the subsequent reanalysis. Intended for
+ * non-UI rapid-edit workflows such as agentic coding tools.
+ *
+ * @default false
+ */
+diagnostics.eagerClear: boolean;
+/**
  * Enable/disable semantic checking of JavaScript files. Existing `jsconfig.json` or `tsconfig.json` files override this setting.
  *
  * @default false
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -226,14 +226,7 @@ completions.completeFunctionCalls: boolean;
 // Diagnostics code to be omitted when reporting diagnostics.
 // See https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json for a full list of valid codes.
 diagnostics.ignoredCodes: number[];
-/**
- * When enabled, diagnostics for a file are cleared immediately on each change,
- * before fresh diagnostics arrive from tsserver. This prevents stale diagnostics
- * from persisting between an edit and the subsequent reanalysis. Intended for
- * non-UI rapid-edit workflows such as agentic coding tools.
- *
- * @default false
- */
+// Clear diagnostics immediately on each change to prevent stale diagnostics from persisting between edits. Default: false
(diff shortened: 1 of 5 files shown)


rchl:
I feel like this is just shifting the problem somewhere else. Not showing diagnostics when the code is broken is potentially as wrong as showing stale diagnostics. A change like this could in theory make agents think that the code is correct when it isn't. So I don't think it fixes the root issue.

I've seen PRs for other projects and looks like you are not having that much luck there either so perhaps it's just not a good idea. I would agree with some other comment that if agents want to really know if the code is valid they should not rely on LSP diagnostics and instead run a full project check.

alliprice:
I get that this feels like it shouldn't be the server's problem. But the server is the only place it can actually be fixed. The agent has no way to know which diagnostics are stale. Only the server knows that its published diagnostics no longer reflect the file it's been told about. The issue is how agents consume diagnostics. An agent's context is a conversational log. It edits a file, then reads whatever the server publishes next. If the server says "3 errors" 0ms after the edit, the agent can't distinguish stale from fresh. It spends tokens analyzing errors that don't exist, maybe reverting a correct fix, then the real diagnostics arrive showing 0 errors. The fix was fine all along.

The concern about agents seeing empty diagnostics and thinking "code is correct" doesn't match how this works in practice. The agent initiated the edit. It knows results are pending. An empty set right after its own edit doesn't fool it into thinking the code is perfect. But stale errors from before the edit absolutely fool it into thinking its fix is broken.

On running a full project check instead, there's a time penalty. You implement your entire fix and then run tests at the end. You find out that you built code on five wrong assumptions that you would have caught in the beginning if you got valid diagnostics. And I just hate the idea of throwing away such a valuable tool when such a small fix would make the behavior compatible for everyone.

PR #1114, opened 2026-07-21
docs: pinning TypeScript version to 6 in the installation instructions

alberic89:
It took me hours to finally find out that I had installed a too recent version of typescript.
This is going to save a lot of time to a lot of noobs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Currently Microsoft is working on [TypeScript 7](https://github.com/microsoft/ty
 ## Installing
 
 ```sh
-npm install -g typescript-language-server typescript
+npm install -g typescript-language-server typescript@6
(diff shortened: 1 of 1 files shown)


rchl:
Make sense but did you see some error from typescript-language-server saying it can't fix any typescript? I think it should have shown a message like:

> Could not find a valid TypeScript installation. Please ensure that the "typescript" dependency is installed in the workspace or that a valid `tsserver.path` is specified. Exiting.

But how that is surfaced can differ depending on LSP client implementation.

alberic89:
Yes, I did see this message, but it puzzled me because I had just installed the typescript package, so I tried to fix my PATH instead of checking what I had really installed.

